PR #13, opened 2018-06-07
Adds some translations

sgiehl:


sgiehl:
will merge this now and update the translations again in a week or so

PR #20, opened 2018-06-14
Add Facebook Pixel Tag Type to Matomo Tag Manager

danielmcclure:
Attempting to create an interface for adding Facebook Pixels Tags. Based on default Google Analytics tag and proposed Google AdWords Tag. 

Note: Tested on a live server and it doesn’t appear to add the section of Tags available? I know this product is new so documentation is light but I’m wondering if there is a missing ingredient for it to show up?

Disclaimer: "Facebook" is a registered trademark of Facebook, Inc. I am not officially partnered with, authorized, endorsed, or sponsored by Facebook. I believe this falls under “General” acceptable use as defined in [Facebook Brand Guidelines](https://en.facebookbrand.com/guidelines/brand) but I’m happy to remove/modify if not.

### DIFF
--- a/Template/Tag/FacebookPixelTag.php
+++ b/Template/Tag/FacebookPixelTag.php
@@ -34,6 +34,11 @@ class FacebookPixelTag extends BaseTag
         return self::CATEGORY_ANALYTICS;
     }
 
+    public function getIcon()
+    {
+        return 'plugins/TagManager/images/icons/F_icon.svg';
+    }
+
     public function getParameters()
     {
         return array(
--- a/Template/Tag/FacebookPixelTag.php
+++ b/Template/Tag/FacebookPixelTag.php
@@ -29,7 +29,7 @@ class FacebookPixelTag extends BaseTag
 
     public function getDescription()
     {
-        return 'The Facebook Pixel is a web analytics and avertising service offered by Facebook.';
+        return 'The Facebook Pixel is a web analytics and advertising service offered by Facebook.';
     }
 
     public function getHelp()
--- a/Template/Tag/FacebookPixelTag.php
+++ b/Template/Tag/FacebookPixelTag.php
@@ -42,15 +42,10 @@ class FacebookPixelTag extends BaseTag
     public function getParameters()
     {
         return array(
-            $this->makeSetting('pixelId', '', FieldConfig::TYPE_STRING, function (FieldConfig $field) {
+            $this->makeSetting('pixelId',, FieldConfig::TYPE_STRING, function (FieldConfig $field) {
                 $field->title = 'Pixel ID';
                 $field->description = 'Long numerical number for Facebook Pixel Code"';
-                $field->validators[] = new NotEmpty();
-                $field->validate = function ($value) {
-                    if (!preg_match('^\d+$', strval($value))) {
-                        throw new \Exception('The Pixel ID should only contain a series of numbers.');
-                    }
-                };
+                $field->validators[] = new NumberRange($min = 0);
             }),
             $this->makeSetting('trackingType', 'pageview', FieldConfig::TYPE_STRING, function (FieldConfig $field) {
                 $field->title = 'Tracking Type';
--- a/Template/Tag/FacebookPixelTag.php
+++ b/Template/Tag/FacebookPixelTag.php
@@ -10,11 +10,19 @@ namespace Piwik\Plugins\TagManager\Template\Tag;
 use Piwik\Settings\FieldConfig;
 use Piwik\Plugins\TagManager\Template\Tag\BaseTag;
 use Piwik\Validators\NotEmpty;
+use Piwik\Validators\NumberRange;
 
 class FacebookPixelTag extends BaseTag
 {
 
-    Public function getName()
+    const ID = 'FacebookPixel';
+
+    public function getId()
+    {
+        return self::ID;
+    }
+
+    public function getName()
     {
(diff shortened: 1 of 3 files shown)


Findus23:
Hi,

When testing your pull request I got the following error:
> Fatal error: Namespace declaration statement has to be the very first statement or after any declare call in the script in /home/lukas/public_html/piwik/plugins/TagManager/Template/Tag/FacebookPixelTag.php on line 9

A closer look shows that you are sending a newline before the PHP script which causes the namespace to not be at the beginning.

I have removed the line in https://github.com/matomo-org/tag-manager/pull/20/commits/a767ed0f5d4756f5f252efdfe954516bd1a0ef0c and this fixed the issue.

Template/Tag/FacebookPixelTag.php
```diff
@@ -0,0 +1,62 @@
...
+    {
+        return self::CATEGORY_ANALYTICS;
+    }
+
+    public function getParameters()
+    {
+        return array(
+            $this->makeSetting('pixelId', '', FieldConfig::TYPE_STRING, function (FieldConfig $field) {
+                $field->title = 'Pixel ID';
+                $field->description = 'Long numerical number for Facebook Pixel Code"';
```
Findus23: I think the `"` is too much here.

Template/Tag/FacebookPixelTag.php
```diff
@@ -0,0 +1,62 @@
...
+                $field->validators[] = new NotEmpty();
+                $field->validate = function ($value) {
```
Findus23: I think you can replace this with a `new NumberRange();` (from `Piwik\Validators\NumberRange`). 
That way the message would also be localized.

danielmcclure: I'm not sure on the minimum/maximum number of digits so think the changes I've made should do for now.

Findus23: The parameter is optional, so this should be good enough.

Template/Tag/FacebookPixelTag.php
Findus23: Ah, sorry that I wrote it that vague. The two single quotes are okay (without them [it's invalid](https://travis-ci.org/matomo-org/tag-manager/jobs/393000757#L913)), but in the line 47 you got an double quote after `Code"` to much.

Template/Tag/FacebookPixelTag.php
```diff
@@ -0,0 +1,62 @@
...
+
+    public function getIcon()
+    {
+        return 'plugins/TagManager/images/icons/F_icon.svg';
+    }
+
+    public function getParameters()
+    {
+        return array(
+            $this->makeSetting('pixelId',, FieldConfig::TYPE_STRING, function (FieldConfig $field) {
```
tsteur: fyi: there are 2 commas... maybe you meant to use `,'',` to have an empty string as default value? If you want to request a number, you could set `FieldConfig::TYPE_STRING` to an integer `FieldConfig::TYPE_INT`

Template/Tag/FacebookPixelTag.web.js
```diff
@@ -0,0 +1,27 @@
...
+                n.queue=[];t=b.createElement(e);t.async=!0;
+                t.src=v;s=b.getElementsByTagName(e)[0];
+                s.parentNode.insertBefore(t,s)}(parameters.window,parameters.document,'script',
+                'https://connect.facebook.net/en_US/fbevents.js');
+            }
+
+            var pixelId = parameters.get('pixelId');
+            if (!(pixelId in setup)) {
+                setup[pixelId] = true;
+                fbq('init', parameters.get('pixelId'));
```
tsteur: just little detail... you may want to re-use `pixelId` instead of calling `parameters.get('pixelId')` again.

tsteur:
Are there any facebook users that can confirm it works?

Template/Tag/FacebookPixelTag.php
danielmcclure: Re-added quotes, this was based on a prior suggestion to remove but seems they are needed to validate. Also updated to TYPE_INT

Template/Tag/FacebookPixelTag.web.js
danielmcclure: Thanks, updated.

tsteur:
Cheers 👍  I can add translations for it later. 

We are checking what is best re the logo, we should add likely an entry in the legalnotice, and maybe even a comment in the file again or so. I'll get back to you on this.

tsteur:
Could we please add a comment within the svg saying something like:

```
Source: https://commons.wikimedia.org/wiki/File:F_icon.svg (public domain)
Facebook is a registered Trademark and this logo may be protected. For more information see https://en.facebookbrand.com/
```

danielmcclure:
@tsteur: I have added your quote in the SVG file comments. I actually reached out to Facebook via their Brand contact form on June 15th for further clarification on brand usage within third party software, specifically relating to this use case. The auto-response advised I should receive a response within one week so we could hold the trigger on publishing anything further until June 22nd and see if there are any further suggestions.

tsteur:
By mentioning the source and the reference to Facebook I would say it should be all fine and we can always change it in case they come back to you later. I can merge the tag later 👍  Kudos on adding this

tsteur:
Kudos @danielmcclure for contributing this tag. Awesome to have this tag 👍

PR #32, opened 2018-07-01
First init triggers, then the dataLayer

tsteur:
Tried to fix https://forum.matomo.org/t/custom-event-trigger-with-mtm-doesnt-work/28993/13 but turns out it is not that trivial to fix. We won't be merging this PR very very likely.

While with this change we would fix the problem in the forum post, it could cause other side effects. Eg using a "pageview" trigger together with a condition of a variable that relies on the data of a dataLayer (eg time since load) could break because the dataLayer would be not populated yet when the triggers are running. 

We will need to spend a couple hours or day or so thinking about how to fix this the best way. Eg only executed missed events but init all variables? (doesn't sound quite right), eg init only the first line of the dataLayer which includes the start event? (doesn't sound quite right either, eg this event might be removed and we would still miss user input). So we need to find a way to still pass this event on. Eg populating all data to the global data layer but to the local container data layer only after the container is running... or knowing which events were missed for which trigger etc.

Edit: I'm seeing all events are already passed to the global data layer and only afterwards to the container so the PR could work.


tsteur:
Note: Looking through the variables the only other template that uses `container datalayer` is the `DataLayerVariable`. Here we could probably fallback to the global data layer if  the container data layer is empty (meaning not set up yet, or to put it in different words: the events have not been passed on from the global data layer to the container data layer).

`DataLayerVariable` should ideally read events from the container data layer as it allows us to have multiple containers in one page with different data layers. It'll be rarely used but be good not to always read from global data layer when this feature is needed.

tsteur:
Actually, I made some more changes and some more testing and this logic doesn't quite work. Noticed it when making the debugger show correct values for when an event is being triggered before the container data layer has been populated. The problem is we would show an empty data layer in the debugger even though the data layer was only semi empty.

This made me think of another problem that we would mess up the data layer entirely by applying this fix as we would first init all triggers. These triggers would populate values to the data layer, which would then be possibly overwritten by initializing the data layer after setting up all triggers. 

So we need to find a different solution.

tsteur:
Fix should work now. Found a way to fix it without having to change behaviour in tagmanager.js

Issue #33, opened 2018-07-04
Document best practice for loading script in tag


Findus23:
I am planning to add a few tags in the next days and came across a question:

The docs for sentry.io suggest adding it like this:

```html
<script src="https://cdn.ravenjs.com/3.26.2/raven.min.js" crossorigin="anonymous"></script>
<script>Raven.config('https://<key>@sentry.io/<project>').install();</script>
```

I guess this is quite common for tags, so how do you think should the JS file be loaded? Like this?

```js
var script = document.createElement('script');
script.onload = function () {
    //do stuff with the script
};
script.src = something;

document.head.appendChild(script);
```

<sub>Thinking about it a sentry.io tag may not be that useful as for some features it needs to be loaded between the library code and application code.</sub>

tsteur:
@Findus23 as this is something that will be used quite often I would add a method to the tag manager library to do this. Thinking:

`TagManager.loadScriptUrl('...', {...someOptions...})`

will create a PR

PR #34, opened 2018-07-05
Add method to insert a JS tag

tsteur:
@Findus23 any thoughts re this API?

refs #33 By default it would be as easy as `TagManager.loadScriptUrl('test.js');` which would be loaded async and deferred. Then various options for different features.


Findus23:
@tsteur Sounds like exactly what I meant. Many thanks.

PR #41, opened 2018-07-25
add Sentry.io tag

Findus23:
Hi,

The first tag should be done. It works in a quick test (even though it should probably only be added on pageload to catch all errors).

The wording can probably be improved a bit, and I am not sure if we maybe should create a new category for such Exception-tracking tools.


Template/Tag/SentryRavenTag.web.js
```diff
@@ -0,0 +1,14 @@
+(function () {
+    return function (parameters, TagManager) {
+        this.fire = function () {
+            var dsn = parameters.get('sentryDSN');
+            if (dsn) {
+                TagManager.dom.loadScriptUrl('https://cdn.ravenjs.com/3.26.4/raven.min.js', {
```
tsteur: @Findus23 I'll merge the PR in a bit... do you know if there is maybe a URL to load the `latest` instead of a specific version?

tsteur:
I think the category works quite well for now, and can re-arrange later if there are more similar ones 👍

PR #42, opened 2018-07-25
Using lookup for matchValue

brosander:


brosander:
@tsteur I had to add this to get value conversion working (basically wanted a catch-all event trigger that wouldn't fire on mtm events

image

tsteur:
Cheers 👍

PR #43, opened 2018-07-25
add twitter conversion tracking

Findus23:
This is just an untested tag. 

Unfortunately the docs only mention copying the tracking code, but without an account one can't see it.

So I took the most basic tracking code I found, but I can't test it.

Here is also another tracking code for tracking a purchase, but I don't know what it can be used for.

https://business.twitter.com/de/help/campaign-measurement-and-analytics/conversion-tracking-for-websites.html

If someone knows more about Twitter conversion tracking, please comment.


Findus23:
According to Google Tag Manager the possible `track`-values are

grafik

but there are no google search results for them and [the twitter docs](https://business.twitter.com/en/help/campaign-measurement-and-analytics/conversion-tracking-for-websites.html) don't mention them. (I can't even get them to display in English instead of German)

tsteur:
What do we need to test it? A twitter business or ad account or something?

Template/Tag/TwitterConversionTag.web.js
```diff
@@ -0,0 +1,17 @@
...
+            var id = parameters.get('Pixel ID');
+            if (id) {
+                !function (e, t, n, s, u, a) {
+                    e.twq || (s = e.twq = function () {
+                        s.exe ? s.exe.apply(s, arguments) : s.queue.push(arguments);
+                    }, s.version = '1.1', s.queue = [], u = t.createElement(n), u.async = !0, u.src = '//static.ads-twitter.com/uwt.js',
+                        a = t.getElementsByTagName(n)[0], a.parentNode.insertBefore(u, a))
+                }(window, document, 'script');
+                twq('init', id);
+                twq('track', 'PageView');
```
tsteur: here in https://business.twitter.com/en/help/campaign-measurement-and-analytics/conversion-tracking-for-websites.html I don't see the conversion type "Pageview"? Only "Site visit", "Purchase", "Download", "Sign up" and "Custom". Also noticed here: https://stackoverflow.com/questions/43937596/are-there-any-different-twitter-pixel-events-than-pageview . 

I'm not sure which parameters each type supports. Only seeing `value`, `num_items` and `order_id`... Their docs are rather not so good.

tsteur:
@Findus23 are you still working on this?

Findus23:
@tsteur No, as neither do I understand the docs nor have a way to test it.

PR #44, opened 2018-07-26
ConstantVariable copy should use a temp variable

brosander:
This prevents the clone from superceding the instance's value


brosander:
@tsteur Matomo Config was always resolving empty strings for data layer params

tsteur:
Good point 👍  It looks like the minified JavaScript is no longer up to date since we merged some other PRs. We can update the minified JS also after merging.

PR #48, opened 2018-08-04
Add Shareaholic Tag

Findus23:
Basically this Tag only adds the `<script type="text/javascript" data-cfasync="false" src="//apps.shareaholic.com/assets/pub/shareaholic.js" data-shr-siteid="[SITEID]" async="async"></script>` that is required for everything they offer to work.

I just tested the most basic feature (added the tag and the share buttons appeared).


tsteur:
Just signed up for this service to test it. I think it would be great if we also supported the possibility to position the share buttons anywhere. It is described in https://support.shareaholic.com/hc/en-us/articles/207375636

and basically to use it, we need to place a div, eg `<div class="shareaholic-canvas" data-app="share_buttons" data-app-id="28214951"></div>` in the page where it should appear. It basically means, we may want to optionally ask for an `App ID` and a selector where to insert this DIV.

Findus23:
I have now added the In-page-apps and they work for me (as in 'successfully make my testpage look like some odd scam site')
grafik

There are a few limitations but I think solving them would make the UI even more complex:

- The Total Share Counter doesn't use an app ID, but rather more data-attributes. So I only support showing the counter for all providers and the current url.
- Only one app can be added at the same time (but more than one of those boxes will make the website look even more weird).

Template/Tag/ShareaholicTag.web.js
```diff
@@ -0,0 +1,41 @@
...
+                    var parentSelector = parameters.get('shareaholicParentSelector');
+                    if (inPageApp !== 'total_share_count' && !AppID) {
+                        TagManager.debug.error('this In-Page App requires an AppID');
+                        return false;
+                    }
+                    if (!parentSelector) {
+                        TagManager.debug.error('you have to specify a Parent Selector to place the In-Page App onto your website');
+                        return false;
+                    }
+                    var parent = TagManager.dom.bySelector(parentSelector)[0];
```
tsteur: I would check that the `parentSelector` was actually found and returns a list. `[0]` may not be defined.

Template/Tag/ShareaholicTag.web.js
```diff
@@ -0,0 +1,41 @@
+(function () {
+    return function (parameters, TagManager) {
+        this.fire = function () {
+            var siteId = parameters.get('shareaholicSiteId');
+            if (siteId) {
+                // should result in a <script> like this:
+                // <script type="text/javascript" data-cfasync="false" src="//apps.shareaholic.com/assets/pub/shareaholic.js" data-shr-siteid="[SITEID]" async="async"></script>
+
+                var script = document.createElement('script');
+                var s = document.getElementsByTagName('script')[0];
```
tsteur: do we need to fallback to `document.body` or `document.head` if no script element can be found?

Template/Tag/ShareaholicTag.php
```diff
@@ -0,0 +1,63 @@
...
+            }),
+
+            $this->makeSetting('shareaholicAppId', '', FieldConfig::TYPE_STRING, function (FieldConfig $field) {
+                $field->title = 'In-Page App ID';
+                $field->description = 'If you want to add an In-Page App enter the App ID here. It is about eight digits long. The App ID is not required for the Total Share Counter.';
+                $field->condition = 'shareaholicInPageApp && shareaholicInPageApp !== "total_share_count"';
+                $field->customUiControlTemplateFile = self::FIELD_TEMPLATE_VARIABLE;
+            }),
+            $this->makeSetting('shareaholicParentSelector', '', FieldConfig::TYPE_STRING, function (FieldConfig $field) {
+                $field->title = 'In-Page App Parent Selector';
```
tsteur: should we validate a parent selector is defined when an app ID is specified? maybe we also need to check that an appId is specified when a parent selector is defined? 

We have something somewhat similar in https://github.com/matomo-org/tag-manager/blob/0.2.0/Template/Variable/DomElementVariable.php#L22-L33 and then  https://github.com/matomo-org/tag-manager/blob/0.2.0/Template/Variable/DomElementVariable.php#L38-L41

So basically, we can save the setting in a variable, and then use it in the validate function and check if a value is set in the other field. There may be one caveat: You can probably only access the value of a previous setting. You may not be able to get the value of `shareaholicParentSelector` while validating `shareaholicInPageApp` but the other way around it should work for sure. So we could at least verify that if pageapp is specified, then we require a value for parent selector.

Template/Tag/ShareaholicTag.web.js
Findus23: The Matomo tracking code does it the same (but is also a `<script>` itself). But I guess it is okay, as using MTM without any script tag sounds unlikely (let's hope no one uses only a `onload=""`)

tsteur: 👍

tsteur:
@Findus23 PR is still set to "WIP". Are you still working on it or waiting for something maybe?

Findus23:
It’s done from my side. 
There are some features that could be added, but I’d wait until the first people start using the tag. (Same is valid for most tags)

Lukas

tsteur:
> There are some features that could be added, but I’d wait until the first people start using the tag. 

Definitely best strategy

jmeattle:
@tsteur @Findus23 thanks so much for adding Shareaholic :)

PR #55, opened 2018-08-10
add bugsnag tag

Findus23:
According to https://trends.builtwith.com/analytics/error-tracking bugsnag is the second most used exeption tracking tool after sentry.io, so let's add a tag.

Again this is untested, but quite straightforward.

## TODO:

- [ ] test if it works
- [ ] improve docs
- [ ] add more options

For the beginning I only support the option to disable tracking of ip addresses, but more options could be added.


Findus23:
@tsteur Thanks for the reminder. Apart from a typo, everything worked fine when I just signed up for a test account.

PR #56, opened 2018-08-10
add tawk.to tag

Findus23:
## TODO:
- [x] find a better category than `CATEGORY_OTHERS`.
- [x] test on real website (I only made a quick check)
- [x] improve docs


tsteur:
What about category "Messaging" or "Communication" but think for now "Social" works quite well too

Findus23:
There are more chat integrations to come so a separate category for them would be great. 

BTW: tomorrow’s a holiday, so I’ll have much time to check for more tags and fix the existing ones. 

Lukas

tsteur:
FYI changed the category and tested it on a website plus tweaked the description to easier find the ID. Worked 👍

PR #58, opened 2018-08-15
add Zendesk Chat Tag

Findus23:
Works in my quick test and is pretty straight forward. 

Unfortunately there is no nice way to copy the ID.

The widget looks like this:
```html
<!--Start of Zendesk Chat Script-->
<script type="text/javascript">
    window.$zopim||(function(d,s){var z=$zopim=function(c){z._.push(c)},$=z.s=
        d.createElement(s),e=d.getElementsByTagName(s)[0];z.set=function(o){z.set.
    _.push(o)};z._=[];z.set._=[];$.async=!0;$.setAttribute("charset","utf-8");
        $.src="https://v2.zopim.com/?5vsZG8Jx0J9FS85qxc6Xn5MMIfVBDzML";z.t=+new Date;$.
            type="text/javascript";e.parentNode.insertBefore($,e)})(document,"script");
</script>
<!--End of Zendesk Chat Script-->
```

I tried to de-minimize it a bit to make it more readable.

### DIFF
--- a/Template/Tag/ZendeskChatTag.web.js
+++ b/Template/Tag/ZendeskChatTag.web.js
@@ -1,0 +1,25 @@
+(function () {
+    return function (parameters, TagManager) {
+        this.fire = function () {
+            var zendeskChatId = parameters.get("zendeskChatId");
+            window.$zopim || (function () {
+                var z = $zopim = function (c) {
+                    z._.push(c)
+                };
+                var script = z.s = document.createElement("script");
+                var e = document.getElementsByTagName("script")[0];
+                z.set = function (o) {
+                    z.set._.push(o)
+                };
+                z._ = [];
+                z.set._ = [];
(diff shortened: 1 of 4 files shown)


Template/Tag/ZendeskChatTag.web.js
```diff
@@ -0,0 +1,25 @@
+(function () {
+    return function (parameters, TagManager) {
+        this.fire = function () {
+            var zendeskChatId = parameters.get("zendeskChatId");
+            window.$zopim || (function () {
```
tsteur: actually, maybe we could add a check to execute it only if a zendeskChatId exists: `if (! zendeskChatId) {return;}`

Findus23: While a check never hurts, I can't think of a way it could be undefined if it is required.

PR #59, opened 2018-08-15
add basic AddThis tag

Findus23:
Pretty similar to the Shareaholic #48 tag.

A bit more simple as everything get's configured via the web dashboard.

### DIFF
--- a/Template/Tag/AddThisTag.web.js
+++ b/Template/Tag/AddThisTag.web.js
@@ -1,6 +1,16 @@
 (function () {
     return function (parameters, TagManager) {
         this.fire = function () {
+            var parentSelector = parameters.get('AddThisParentSelector');
+            if (parentSelector) {
+                var parent = TagManager.dom.bySelector(parentSelector);
+                if (parent) {
+                    var div = document.createElement('div');
+                    div.className = 'addthis_inline_share_toolbox';
+                    parent[0].appendChild(div)
+                }
+            }
+
             var siteId = parameters.get('AddThisPubId');
             if (siteId) {
                 var script = document.createElement('script');
@@ -8,16 +18,6 @@
                 script.type = 'text/javascript';
                 script.src = '//s7.addthis.com/js/300/addthis_widget.js#pubid=' + siteId;
                 s.parentNode.insertBefore(script, s);
-
-                var parentSelector = parameters.get('AddThisParentSelector');
-                if (parentSelector) {
-                    var parent = TagManager.dom.bySelector(parentSelector);
-                    if (parent) {
-                        var div = document.createElement('div');
-                        div.className = 'addthis_inline_share_toolbox';
-                        parent[0].appendChild(div)
-                    }
-                }
             }
         };
     };
--- a/Template/Tag/AddThisTag.web.js
+++ b/Template/Tag/AddThisTag.web.js
@@ -1,0 +1,22 @@
+(function () {
+    return function (parameters, TagManager) {
+        this.fire = function () {
+            var siteId = parameters.get('AddThisPubId');
+            if (siteId) {
+                var script = document.createElement('script');
+                var s = document.getElementsByTagName('script')[0];
+                script.type = 'text/javascript';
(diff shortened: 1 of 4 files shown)


Template/Tag/AddThisTag.web.js
```diff
@@ -0,0 +1,22 @@
...
+            if (siteId) {
+                var script = document.createElement('script');
+                var s = document.getElementsByTagName('script')[0];
+                script.type = 'text/javascript';
+                script.src = '//s7.addthis.com/js/300/addthis_widget.js#pubid=' + siteId;
+                s.parentNode.insertBefore(script, s);
+
+                var parentSelector = parameters.get('AddThisParentSelector');
+                if (parentSelector) {
+                    var parent = TagManager.dom.bySelector(parentSelector)[0];
```
tsteur: same here, we should make sure that a resultSet was found before accessing `[]`. 

Same above  in `document.getElementsByTagName('script')[0]` where we maybe need to fallback to `document.body`? Although now that I think about it: This check here may not be needed as the script element for the tag manager should exist anyway and it should always find one script element.

Findus23: Fixed selector in https://github.com/matomo-org/tag-manager/pull/59/commits/96f6cedcf592543699246f2ea9572b86b630c3d1

Template/Tag/AddThisTag.web.js
```diff
@@ -0,0 +1,24 @@
...
+        this.fire = function () {
+            var siteId = parameters.get('AddThisPubId');
+            if (siteId) {
+                var script = document.createElement('script');
+                var s = document.getElementsByTagName('script')[0];
+                script.type = 'text/javascript';
+                script.src = '//s7.addthis.com/js/300/addthis_widget.js#pubid=' + siteId;
+                s.parentNode.insertBefore(script, s);
+
+                var parentSelector = parameters.get('AddThisParentSelector');
```
tsteur: I wonder should this maybe executed before inserting the script? (haven't tested it but wondering how the script above would know this DOM was inserted afterwards).  I presume so far it works because it'll take a while to load the script above.

Template/Tag/AddThisTag.web.js
```diff
@@ -0,0 +1,24 @@
+(function () {
+    return function (parameters, TagManager) {
+        this.fire = function () {
+            var siteId = parameters.get('AddThisPubId');
+            if (siteId) {
+                var script = document.createElement('script');
+                var s = document.getElementsByTagName('script')[0];
+                script.type = 'text/javascript';
+                script.src = '//s7.addthis.com/js/300/addthis_widget.js#pubid=' + siteId;
```
tsteur: Probably won't make a big difference but could potentially load it async and deferred (I reckon current implementation is what addthis suggests)

Template/Tag/AddThisTag.web.js
Findus23: It definitely doesn't hurt doing it first.

Template/Tag/AddThisTag.web.js
Findus23: I'm not sure if it changes anything as it only gets executed after pageload. But I guess we can add it to all tags (maybe except for error tracking).

lang/en.json
```diff
@@ -1,5 +1,7 @@
 {
    "TagManager": {
+      "AddThisTagDescription": "Add the AddThis social media buttons and more to your website.",
+      "AddThisTagHelp": "TODO",
```
tsteur: should we just not define a help text? eg in the tag we can add:

```
public function getHelp()
    {
return '';
}
```

Findus23: I am not sure what we could write as a good help text for AddThis.

tsteur: That's why I meant we could just leave it empty. The description is already pretty clear.

Findus23: Makes sense. Done in https://github.com/matomo-org/tag-manager/pull/59/commits/bd9ecec86d6fb60ae1ba075c10bb82408922eab7

PR #65, opened 2018-08-24
Added Emarsys tag

MichaelHeerklotz:
Basic Emarsys support, currently with these commands:  category, view, tag, go
Use case: landing pages or similar that supplement a given shop system


Template/Tag/EmarsysTag.php
```diff
@@ -0,0 +1,75 @@
...
+            }),
+            $this->makeSetting('commandGo', '', FieldConfig::TYPE_BOOL, function (FieldConfig $field) {
+                $field->title = 'go';
+                $field->uiControl = FieldConfig::UI_CONTROL_CHECKBOX;
+                $field->description = 'Execute commands in the queue, that is, send them to the recommender service for processing.';
+            }),
+        );
+    }
+
+    public function getOrder()
```
Findus23: Until we have a significant number of different tags, I'd leave the default order and maybe order all of them afterwards.

MichaelHeerklotz:
Thank you for your feedback and help, I have applied the changes.

PR #76, opened 2018-09-08
add Bing Conversion Tracking

Findus23:
Not tested

The original tracking code is 
```html
<script>(function(w,d,t,r,u){var f,n,i;w[u]=w[u]||[],f=function(){var o={ti:"26017395"};o.q=w[u],w[u]=new UET(o),w[u].push("pageLoad")},n=d.createElement(t),n.src=r,n.async=1,n.onload=n.onreadystatechange=function(){var s=this.readyState;s&&s!=="loaded"&&s!=="complete"||(f(),n.onload=n.onreadystatechange=null)},i=d.getElementsByTagName(t)[0],i.parentNode.insertBefore(n,i)})(window,document,"script","//bat.bing.com/bat.js","uetq");</script>
```


Template/Tag/BingUETTag.php
```diff
@@ -0,0 +1,39 @@
...
+        return 'plugins/TagManager/images/icons/bing.svg';
+    }
+
+    public function getParameters()
+    {
+        return array(
+            $this->makeSetting('bingAdID', '', FieldConfig::TYPE_STRING, function (FieldConfig $field) {
+                $field->title = 'ID';
+                $field->uiControl = FieldConfig::UI_CONTROL_TEXT;
+                $field->description = '';
```
tsteur: Is there an easy way to get the ID? PR looks good otherwise (haven't tested though)

Findus23: Ah, I forgot the description, because I couldn't think of a great wording.

The ID is quite visible in the table where new tags can be created and managed and is also displayed before the tracking code in the modal.

PR #77, opened 2018-09-10
Move UI test screenshots to Git LFS

sgiehl:


sgiehl:
Small side note: Seems the other plugins included in core, don't store their UI files on LFS, so it might need adjustments in core for the tests to pass as I'm not sure if LFS files are fetched on a submodule update...

Issue #79, opened 2018-09-19
Add "Match CSS Selector" trigger condition


ben-dw:
Hi there, 
how can I use the variable "DomElementVariable". Do we have to execute the command 

`$ ./console generate:tagmanager-preconfigured-variable`
?
Thank you for helping out!

ben-dw:
Sorry for the confusion.
Yes, I'd like to use it as a trigger, like in GTM.
In GTM it is possible to choose "Click Element" using CSS selector to get ".class1 .class2". Is there a similar way in matomo?

tsteur:
Yes it works pretty much the same way. Just go to the trigger you want to use and configure the variable where you need it.

ben-dw:
But is there a way to use a css selector instead of defining x variables like in GTM?
![image](https://user-images.githubusercontent.com/43407231/45818647-a152f780-bce2-11e8-954e-7f393468d42e.png)

tsteur:
I see, now I get it. Didn't know they had this feature :) 

We could add a new comparsion "Matches CSS selector" in https://github.com/matomo-org/tag-manager/blob/0.2.3/Model/Comparison.php#L39 and process it in https://github.com/matomo-org/tag-manager/blob/0.2.3/javascripts/tagmanager.js#L202 

Ideally we would have a CSS selector validator when entering a CSS selector but this is not needed for now. We will add this at a later point and also add it then to other places where a user can enter a CSS selector.

How does this feature work? Does it only work for elements, eg `ClickElement`?

ben-dw:
Awesome!
It works for `ClickElement` and `FormElement`.

tsteur:
So far I wasn't pushing the `ClickElement` in click events as I was afraid of possible memory leaks. The DOM node would be kept in the tag manager as a reference. Would maybe need to see if we could add an API that lets it delete the reference after the conditions have been evaluated.

mattab:
Documented in [ How do I trigger a tag when any DOM element which matches a CSS selector is clicked?](https://matomo.org/faq/tag-manager/how-do-i-trigger-a-tag-when-any-dom-element-matches-a-css-selector/)

PR #83, opened 2018-09-30
Add support for Click Element Match CSS Selector

tsteur:
fix https://github.com/matomo-org/tag-manager/issues/79

note: when an element is clicked, we keep a reference to this node in all the dataLayers. This is likely not a big problem for the link and download click handler but could cause memory issues for the "all elements click handler". If/when it becomes a problem, could use eg `MutationObserver` to monitor when it is removed from DOM and unset the memory. So far we need to keep an instance of all events in case a container is loaded at any time in which case we would need to replay all previous events.


mattab:
Documented in [ How do I trigger a tag when any DOM element which matches a CSS selector is clicked?](https://matomo.org/faq/tag-manager/how-do-i-trigger-a-tag-when-any-dom-element-matches-a-css-selector/)

PR #106, opened 2018-10-06
Copy trigger

### LINKED ISSUE #87: New trigger: Copy text
It could be valuable to have this trigger as it allows you to for example track an event when someone copies something on your website which basically is an interaction. It helps you get more realistic bounce rate for example.

Findus23:
fixes #87  
@tsteur should I also add `copyElementId`-like variables similar to the copy target?


Template/Trigger/CopyTrigger.web.js
```diff
@@ -0,0 +1,20 @@
...
+        this.setUp = function (triggerEvent) {
+            TagManager.dom.addEventListener(parameters.document, "copy", function (event) {
+                if (!event.target) {
+                    return;
+                }
+                var target = event.target;
+                var dom = TagManager.dom;
+                triggerEvent({
+                    event: 'mtm.Print',
+                    'mtm.copyElement': target,
```
tsteur: @Findus23 could we also post the copied text and the text length similar to https://gist.github.com/ju1/6fb3294d28a1461e647349c69a4e30e3#file-copytextrecipe-json-L101 ? The length allows users to filter a specific length like > 30 & < 300 characters which might be eg copied for a tweet.

Template/Trigger/CopyTrigger.web.js
```diff
@@ -0,0 +1,20 @@
...
+    return function (parameters, TagManager) {
+        this.setUp = function (triggerEvent) {
+            TagManager.dom.addEventListener(parameters.document, "copy", function (event) {
+                if (!event.target) {
+                    return;
+                }
+                var target = event.target;
+                var dom = TagManager.dom;
+                triggerEvent({
+                    event: 'mtm.Print',
```
tsteur: The event name be "Copy"

tsteur:
Sure sounds good 👍 Might be enough to have it only for copy text and copy length but others could be added too

Findus23:
Copy Length sounds good, but I am really not sure if we should add a copy text variable as it makes it really easy to do stupid things.

BTW: Is there a reason why the en.json is indented with 3 spaces instead of 4? (Because at the moment the `generate:` tool always sets it back to 4)

tsteur:
> BTW: Is there a reason why the en.json is indented with 3 spaces instead of 4? (Because at the moment the generate: tool always sets it back to 4)

No, that must be definitely a mistake. 

> Copy Length sounds good, but I am really not sure if we should add a copy text variable as it makes it really easy to do stupid things.

I would say people can do it potentially anyway but it makes it certainly easier. Similar to dom element variable it should not do it for passwords: https://github.com/matomo-org/tag-manager/blob/master/Template/Variable/DomElementVariable.web.js#L26-L29  Could also eventually support some attributes from Heatmap & Session Recording for those variables and include `data-matomo-mask` etc. We could also disable it actually in general for any form field.

Template/Trigger/CopyTrigger.web.js
```diff
@@ -0,0 +1,31 @@
...
+                var selectedText;
+                if (window.getSelection) {
+                    selectedText = window.getSelection().toString();
+                } else if (document.selection && document.selection.type !== 'Control') {
+                    selectedText = document.selection.createRange().text;
+                } else {
+                    TagManager.debug.error("can't detect selected text");
+                    return false;
+                }
+                console.info(selectedText);This trigger is not supported in Internet Explorer
```
tsteur: fyi there is also still a console in here

tsteur:
I did a quick search but couldn't find what the problems are. People that use the copy trigger for tracking also don't mention anything. Do you know what exactly the cross browser problems are?

sgiehl:
@tsteur @Findus23  is that something that should still be finished and merged?

tsteur:
We could close it for now. Eventually it be great to have but not too important.

PR #123, opened 2018-11-05
Support custom title and url for pageviews

### LINKED ISSUE #122: Matomo tag: set a custom page name when tracking a pageview
When tracking  a pageview in Matomo, sometimes it is useful to overwrite the page title and set a custom page title. 

This is done via the JS function `setDocumentTitle`.

## workaround available?

When tracking a pageview via the Tag Manager, and if you need to set a custom page name, currently it may be possible (to be confirmed it works) to create a Custom html tag with the following code:
```js
    window.piwikPluginAsyncInit.push(function () {
        window.Piwik.on('TrackerSetup', function (tracker) {
            tracker.setDocumentTitle('My New Title');
       });
   });
```

tsteur:
fix #122 

@mattab can you give this a try?

image

### DIFF
--- a/Template/Tag/MatomoTag.php
+++ b/Template/Tag/MatomoTag.php
@@ -66,6 +66,24 @@ class MatomoTag extends BaseTag
                     $field->validators[] = new CharacterLength(1, 500);
                 }
             }),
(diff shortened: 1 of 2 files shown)


jmumby:
I set this up using the page url and local time as variables. Works perfectly

tsteur:
Cheers @jmumby

tsteur:
Will need to fix tests and then can be merged

PR #141, opened 2018-12-15
add Fathom tag

Findus23:
[Fathom](https://usefathom.com/) is an open source, self-hostable analytics solution and therefore should definitely needs its own tag in MTM.

While using the tracking code directly works, with tag manager there doesn't seem to be a pageview tracked (even though calling `fathom('trackPageview');` manually works)
I am not sure why this doesn't work, maybe @dannyvankooten has an idea.


Template/Tag/FathomTag.web.js
```diff
@@ -0,0 +1,26 @@
+(function () {
+    return function (parameters, TagManager) {
+        this.fire = function () {
+            var siteID = parameters.get('siteID');
+            var fathomURL = parameters.get('fathomURL');
+            if (siteID && fathomURL) {
+                (function () {
+                    window.fathom = window.fathom || function () {
```
dannyvankooten: I would check if `window.fathom` exists here and bail if it does, to prevent adding the `<script>` tag to the DOM multiple times.

Findus23: Thanks this makes sense.

tsteur:
@Findus23 do you know what the status is here?

Findus23:
@tsteur When I first tested it, it didn't work and I didn't get around to checking it again.

tsteur:
@Findus23 closing this for now. If someone is keen to get Fathom working feel free to reopen the PR or  reuse the code as a start.

PR #150, opened 2019-01-18
Add random id to preview url to avoid browser cache

tsteur:
Avoids having to force reload with clearing browser cache if preview changes. Only problem may be debugging is that debugging in the browser might be harder. Could be interesting having an option not to append the parameter.


tsteur:
Created https://github.com/matomo-org/matomo/pull/13977 instead

PR #158, opened 2019-03-29
Matomo Tag: set the id site later so that cookies are created correctly

mattab:
Steps to reproduce:
* Create a Matomo tag on a container, configured with cookie domain `.example.com`
* Load the website www.example.com
* Got: Two cookies are created on domains .example.com and www.example.com eg.
```
9b1feac345884afb.1553796770.0.1553796834.. on domaine www.example.com
97b1f55eea218e3e.1553796770.1.1553796834.1553796770. on .example.com
```
* Expected only one cookie on .example.com

Looked a bit into it and found that:
* in the container it calls `        tracker = Piwik.addTracker(trackerUrl, matomoConfig.idSite);`
* which then triggers the line `setVisitorIdCookie();` in the Tracker constructor
* which creates the wrong cookie on `www.example.com`
* the wrong cookie is created because at this point, the function `setCookieDomains` hasn't been called yet. It's called a bit later in: https://github.com/matomo-org/tag-manager/blob/0.2.6/Template/Tag/MatomoTag.web.js#L121-L123
* hence why in this PR by moving the setSiteId to after the `setCookie*` functions are called, I think may fix the issue

Note: it is un-tested

### DIFF
--- a/Template/Tag/MatomoTag.web.js
+++ b/Template/Tag/MatomoTag.web.js
@@ -126,6 +126,8 @@
                         tracker.setCookiePath(matomoConfig.cookiePath);
                     }
 
+                    tracker.setSiteId(matomoConfig.idSite);
+                    
                     if (matomoConfig.alwaysUseSendBeacon) {
                         tracker.alwaysUseSendBeacon();
                     }
@@ -148,8 +150,6 @@
                     if (matomoConfig.trackVisibleContentImpressions) {
                         tracker.trackVisibleContentImpressions();
                     }
-                    
-                    tracker.setSiteId(matomoConfig.idSite);
                 }
 
                 if ((matomoConfig.userId || tracker.getUserId()) && lastUserId !== matomoConfig.userId) {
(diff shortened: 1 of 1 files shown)


Template/Tag/MatomoTag.web.js
```diff
@@ -148,6 +148,8 @@
                     if (matomoConfig.trackVisibleContentImpressions) {
                         tracker.trackVisibleContentImpressions();
                     }
+                    
+                    tracker.setSiteId(matomoConfig.idSite);
```
mattab: maybe it should be rather just below the call to `setCookiePath` as the functions above (alwaysUseSendBeacon, enableLinkTracking, etc.) may need the id site to function properly.

tsteur: var applyFirst = ['addTracker', 'disableCookies', 'setTrackerUrl', 'setAPIUrl', 'enableCrossDomainLinking', 'setCrossDomainLinkingTimeout', 'setSecureCookie', 'setCookiePath', 'setCookieDomain', 'setDomains', 'setUserId', 'setSiteId', 'alwaysUseSendBeacon', 'enableLinkTracking', 'requireConsent', 'setConsentGiven'];
this is the order

tsteur:
Looks good to merge. In general this should be ideally fixed in core though cause what this is showing is that all users that use `getTracker/addTracker/getAsyncTracker` have problems.

PR #164, opened 2019-05-30
check if element exists

tsteur:
todo: add OnReady to AllClickElementHandler.


mattab:
Customer has confirmed that this fixes the issue @tsteur - nice work 💯

PR #171, opened 2019-07-24
Prevent error "cannot read property 'firstChild' of null"

tsteur:
When Custom HTML tag is set to be inserted into the body, but the tag manager is embedded in the script header and loaded and executed before the body is present in the dom, we need to wait for the body to become available.


tsteur:
Unfortunately not @diosmosis

PR #175, opened 2019-08-15
Fix tests

sgiehl:
* marked `ContextTest::test_generate_web` as skipped for Matomo < 3.8 as output seems to have changed
* update expected system test files
* updated UI files after npm updates in matomo-org/matomo#14680


tsteur:
BTW @sgiehl because we're shipping Tag Manager with Matomo, and not on the Marketplace, does it make sense to remove the `require` here: https://github.com/matomo-org/tag-manager/blob/0.2.6/plugin.json#L7

Considering it will only need to be compatible with the most recent version? Not sure how we handle this usually.

sgiehl:
guess we could even remove the complete `plugin.json` file :thinking: and change tests here to run only with latest release?

tsteur:
Not sure if it works to remove `plugin.json` or whether something will be missing then? I suppose not. Sounds good to make the change 👍  will create a PR and we see if it works

PR #176, opened 2019-08-16
Adjustments for using TM and classic Tracking Code at the same time

MichaelHeerklotz:
* Added new setting "Use external Tracker" that allows to use a Matomo JS Tracker that is already loaded by the website by other means (for example the classic Matomo JS tracking code)
* Replaced Piwk.addTracker() with Piwik.getTracker(); this fixes using Matomo Tag Manager together with classic Matomo Code. addTracker() does some caching and did not work in my tests. getTracker() always creates a new Piwik instance, but since the Tag Manage code already caches the Piwik Instance this should be no problem

### DIFF
--- a/Template/Tag/MatomoTag.web.js
+++ b/Template/Tag/MatomoTag.web.js
@@ -103,7 +103,11 @@
                     // but even two or more different configs for the same Matomo URL & idSite
                     lastMatomoUrl = getMatomoUrlFromConfig(matomoConfig);
                     var trackerUrl = lastMatomoUrl + matomoConfig.trackingEndpoint;
-                    tracker = Piwik.addTracker(trackerUrl);
+                    if (matomoConfig.registerAsDefaultTracker) {
+                        tracker = Piwik.addTracker(trackerUrl);
+                    } else {
+                        tracker = Piwik.getTracker(trackerUrl);
+                    }
                     configuredTrackers[variableName] = tracker;
 
                     if (matomoConfig.disableCookies) {
--- a/Template/Variable/MatomoConfigurationVariable.php
+++ b/Template/Variable/MatomoConfigurationVariable.php
@@ -211,6 +211,11 @@ class MatomoConfigurationVariable extends BaseVariable
                 $field->uiControl = FieldConfig::UI_CONTROL_CHECKBOX;
                 $field->description = 'By bundling the Matomo JavaScript tracker directly into the container it may improve the performance of your website as it reduces the number of needed requests. It is recommended to bundle the Matomo tracker because in most cases the tracker would otherwise be otherwise loaded in a separate request on page view anyway. Note: If you use two different Matomo configurations in one container, the setting of the first configuration used in the first Matomo Tag will be applied to all Matomo tags within one container.';
             }),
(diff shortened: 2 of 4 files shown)


tsteur:
`Piwik.addTracker` had some bug that we'll fix shortly (it required idSite) to be set when it shouldn't. We'll fix that in core.
We don't really want to use `Piwik.getTracker` here as then you wouldn't be able to track events  or other things on the site using `_paq.push(['...'])` anymore. `getTracker` does not register an "async tracker" basically

MichaelHeerklotz:
Actually this is what I am trying to accomplish.
In my setup we have a lot of _paq.push() code that comes along with the classic Matomo Tracking Code. Now we want to add a Matomo Tag,  that should operate independant of the _paq.push() code (different site ids, different events, etc).

Now that you mention it, I see that it can make sense to support using _paq.push() when using Matomo Tag Manager, but I guess not many people are using this kind of setup (mixing tags and custom javascript tracking code). Maybe it makes sense to add an option to disable/enable this.

My users already tried to add the Matomo Tag Manager Js Code into their pages that already have the Classic Js Tracker Code, and assumed that it would work independently.

tsteur:
Setting to separate this behaviour sounds good I reckon.

So we have a setting to use `addTracker(default)` or `getTracker`.
* `addTracker` should be used when any `_paq.push()` in the website should also be applied to this tag.
* `getTracker` should be used when `_paq.push` should not affect the tracker in Matomo Tag.

Be good to explain this "nicely" in an understandable way :)

Template/Tag/MatomoTag.web.js
```diff
@@ -210,6 +210,10 @@
             // to initialize the tracker during window.piwikPluginAsyncInit
 
             var matomoConfig = parameters.get('matomoConfig', {});
+            if (matomoConfig.useExternalTracker) {
```
tsteur: This one I'm not quite sure I understand?

MichaelHeerklotz:
Thank you for your feedback. I have added a new setting "registerAsDefaultTracker" to switch between addTracker() and getTracker().

Template/Tag/MatomoTag.web.js
MichaelHeerklotz: Thank you for your feedback. I have merged this setting with the "bundleTracker" setting into a new "trackerSource" setting. 
trackerSource=external is for websites that already load matomo.js (asynchronously) and do not require the js tracker to be loaded by the container, while the container cannot detect it is already being loaded because it is loaded async.

MichaelHeerklotz: Because the "bundleTracker" setting gets removed with this change, already existing configs will default back to  trackerSource=bundle, if you do not like this, we can change it

MichaelHeerklotz:
I am going to update this pr soon, so no need to merge it right now.

tsteur:
Awesome, let me know when you're done

mattab:
> Piwik.addTracker had some bug that we'll fix shortly (it required idSite) to be set when it shouldn't. 
> We'll fix that in core.
fyi this was fixed in https://github.com/matomo-org/matomo/pull/14787

MichaelHeerklotz:
@tsteur I have now updated the commit to include only the "registerAsDefaultTracker" settings.

I have dropped the other setting ("Use external Tracker") because it did not prove to be a stable way to load the tracker js.

Issue #160, opened 2019-05-02
Add field for setDomains on the Matomo Configuration variable


tomper00:
The need here is to be able to execute the setDomains command in the Matomo tags.
I am not sure if this should be available in the Matomo Configuration variable or if it is something you should set on the Event / Pageview level

This is what we want to be able to do:
// Tell Matomo the website domain so that clicks on these domains are not tracked as 'Outlinks'
_paq.push(['setDomains', '*.example.com']);
_paq.push(['trackPageView']);

PR #177, opened 2019-08-16
Added a config setting for "setDomains"

MichaelHeerklotz:
This setting allows to configure the setDomains() call. Related: https://github.com/matomo-org/tag-manager/issues/160

### DIFF
--- a/Template/Variable/MatomoConfigurationVariable.php
+++ b/Template/Variable/MatomoConfigurationVariable.php
@@ -124,6 +124,39 @@ class MatomoConfigurationVariable extends BaseVariable
                 $field->description = 'When tracking many subdirectories in separate websites, the cookie path prevents the number of cookies to quickly increase and prevent browser from deleting some of the cookies. This ensures optimal data accuracy and improves performance for your users (fewer cookies are sent with each request).';
                 $field->validators[] = new CharacterLength(0, 500);
             }),
+
+            $this->makeSetting('domains', array(), FieldConfig::TYPE_ARRAY, function (FieldConfig $field) {
+                $field->title = 'Domains';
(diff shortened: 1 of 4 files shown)


Template/Variable/MatomoConfigurationVariable.php
```diff
@@ -124,6 +124,13 @@ public function getParameters()
                 $field->description = 'When tracking many subdirectories in separate websites, the cookie path prevents the number of cookies to quickly increase and prevent browser from deleting some of the cookies. This ensures optimal data accuracy and improves performance for your users (fewer cookies are sent with each request).';
                 $field->validators[] = new CharacterLength(0, 500);
             }),
+
+            $this->makeSetting('domains', '', FieldConfig::TYPE_STRING, function (FieldConfig $field) {
+                $field->title = 'Domains';
+                $field->description = 'Set array of hostnames or domains to be treated as local. For wildcard subdomains, you can use: ".example.com" or "*.example.com". You can also specify a path along a domain: "*.example.com/subsite1"';
```
tsteur: Should the description maybe say something like comma separated list? Or how would one actually configure an array? (haven't looked what `setDomains`) supports

tsteur: Just checked `setDomains` expects an array... 
btw be good to have this setting 👍 

I'm not sure if we have a problem or not... haven't tried to implement it and don't know if I remember details 100% correct whether there is a problem when someone wants to define a string to read it from a TagManager Variable or when someone wants to define an array of strings. Not sure if both is actually possible at the same time yet. Not sure you understand.

MichaelHeerklotz: Thank you. The code is now using an array for setDomains().
I have also updated the description a bit, most of it comes from the setDomains() from https://developer.matomo.org/api-reference/tracking-javascript

tsteur:
Cheers 👍  works nicely

PR #178, opened 2019-08-16
Added settings to configure piwik.js and piwik.php path

MichaelHeerklotz:
Added support for js/ and js/tracker.php endpoints.

### DIFF
--- a/Template/Variable/MatomoConfigurationVariable.php
+++ b/Template/Variable/MatomoConfigurationVariable.php
@@ -173,6 +173,30 @@ class MatomoConfigurationVariable extends BaseVariable
                 $field->uiControl = FieldConfig::UI_CONTROL_CHECKBOX;
                 $field->description = 'By bundling the Matomo JavaScript tracker directly into the container it may improve the performance of your website as it reduces the number of needed requests. It is recommended to bundle the Matomo tracker because in most cases the tracker would otherwise be otherwise loaded in a separate request on page view anyway. Note: If you use two different Matomo configurations in one container, the setting of the first configuration used in the first Matomo Tag will be applied to all Matomo tags within one container.';
             }),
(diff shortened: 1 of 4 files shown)


Template/Variable/MatomoConfigurationVariable.php
```diff
@@ -173,6 +173,28 @@ public function getParameters()
                 $field->uiControl = FieldConfig::UI_CONTROL_CHECKBOX;
                 $field->description = 'By bundling the Matomo JavaScript tracker directly into the container it may improve the performance of your website as it reduces the number of needed requests. It is recommended to bundle the Matomo tracker because in most cases the tracker would otherwise be otherwise loaded in a separate request on page view anyway. Note: If you use two different Matomo configurations in one container, the setting of the first configuration used in the first Matomo Tag will be applied to all Matomo tags within one container.';
             }),
+            $this->makeSetting('jsEndpoint', 'piwik.js', FieldConfig::TYPE_STRING, function (FieldConfig $field) {
+                $field->title = 'Tracker Javascript Path';
+                $field->uiControl = FieldConfig::UI_CONTROL_SINGLE_SELECT;
+                $field->availableValues = array(
```
tsteur: we'd also want `matomo.js` and `matomo.php`. Why not in general directly let users enter a value? Some rename them to say `tracker.js` or something else when they want to avoid ad blockers etc.

MichaelHeerklotz: Good point. I have added matomo.js/.php. Having a dropdown seems a bit more user friendly for not so advanced users to me. In most cases using /js/ should be good enough. I did some testing a while ago, and using /js/ for js and tracking requests was not blocked by any browser extension. If you like a free text value better, maybe it still makes sense to keep the dropdown and add something like "other" into it.

tsteur:
Cheers @MichaelHeerklotz

PR #179, opened 2019-08-16
Run tests only against latest version

tsteur:


.travis.yml
```diff
@@ -40,9 +40,6 @@ env:
 matrix:
   exclude:
     # execute latest stable tests only w/ PHP 5.6
-    - php: 5.5
-      env: TEST_SUITE=PluginTests MYSQL_ADAPTER=PDO_MYSQL TEST_AGAINST_CORE=minimum_required_piwik
-    # execute UI tests only w/ PHP 5.6
```
diosmosis: I think this means it will run minimum_required_piwik tests w/ both 5.5 and 5.6 now, did you mean to remove something in the `include:` section?

tsteur: just did i think 👍

PR #186, opened 2019-09-01
Prevent possible circular reference in preview mode

### LINKED ISSUE #185: Tag Manager Preview UI tries to encode circular structure as JSON
Hello,
If the preview UI tries to display a data structure with a circular structure it fails at JSON encoding, an uncaught exception gets thrown and the UI doesn't update.

This can happen for example in the following setting:
- define an `AllElementsClick` trigger
- open the tracked website in debug / preview mode
- give an element on the page a reference to itself, e.g. `el.selfref = el;`
- click that element
- `mtm.clickElement` will be pushed to the datalayer
- And here is what I assume is happening: the preview UI tries to JSON encode the datalayer to display it and fails due to the circular structure caused by the self reference.

Of course the given example is a bit silly, however for example when a UI framework that binds a circular structure to the element is used this becomes an issue.
For instance, vue binds a `__vue__` property to the root element of a component from which a circular structure is given via `__vue__.$el`.

tsteur:
fix #185

Not tested it.


tsteur:
created https://github.com/matomo-org/tag-manager/pull/265 where the base is set to matomo 4

PR #192, opened 2019-10-15
 feat: Created a config key to define custom containers and analytics url 

lockland:
This features wants to provide a way to custom the entrypoint to static
containers and analytics.
In a high traffic machine, analytics will use too much resource and it's
natural to use it behind a load balancer, but provide the static files on
the same serve may take a long time to response impacting on sites
performance.

Config Keys. 
custom_containers_hostname 
custom_analytics_hostname

Example: 
custom_containers_hostname = cdn.example.com 
custom_analytics_hostname = analytics.example.com

### DIFF
--- a/Template/Variable/MatomoConfigurationVariable.php
+++ b/Template/Variable/MatomoConfigurationVariable.php
@@ -9,11 +9,11 @@ namespace Piwik\Plugins\TagManager\Template\Variable;
 
 use Piwik\Common;
 use Piwik\Settings\FieldConfig;
-use Piwik\SettingsPiwik;
 use Piwik\Site;
 use Piwik\Validators\CharacterLength;
 use Piwik\Validators\NotEmpty;
 use Piwik\Validators\UrlLike;
+use Piwik\Plugins\TagManager\SystemSettings as Settings;
 
 class MatomoConfigurationVariable extends BaseVariable
 {
@@ -42,9 +42,10 @@ class MatomoConfigurationVariable extends BaseVariable
     public function getParameters()
     {
         $idSite = Common::getRequestVar('idSite', 0, 'int');
+        $settings = new Settings();
+        $url = $settings->getAnalyticsUrl();
 
-        $url = SettingsPiwik::getPiwikUrl();
-        if (SettingsPiwik::isHttpsForced()) {
+        if ($settings->isHttpsForced()) {
             $url = str_replace('http://', 'https://', $url);
         } else {
             $url = str_replace(array('http://', 'https://'), '//', $url);
--- a/Context/WebContext.php
+++ b/Context/WebContext.php
@@ -22,8 +22,7 @@ use Piwik\Plugins\TagManager\Model\Tag;
 use Piwik\Plugins\TagManager\Model\Trigger;
 use Piwik\Plugins\TagManager\Model\Variable;
 use Piwik\Plugins\TagManager\Template\Variable\VariablesProvider;
-use Piwik\SettingsPiwik;
-
+use Piwik\Plugins\TagManager\SystemSettings as Settings;
 
 class WebContext extends BaseContext
 {
@@ -212,7 +211,7 @@ class WebContext extends BaseContext
 
     private function getWebPathForRelease($idSite, $idContainer, $environment, $containerCreatedDate)
     {
-        $domain = SettingsPiwik::getPiwikUrl();
+        $domain = (new Settings())->getContainersUrl();
         if (Common::stringEndsWith($domain, '/')) {
             $domain = Common::mb_substr($domain, 0, -1);
         }
--- a/SystemSettings.php
+++ b/SystemSettings.php
@@ -175,7 +175,7 @@ class SystemSettings extends \Piwik\Settings\Plugin\SystemSettings
         return $this->replacePiwikUrl($this->customAnalyticsHostname->getValue());
     }
 
-    private function replacePiwikUrl($newUrl): string
+    private function replacePiwikUrl($newUrl)
     {
         $oldUrl = SettingsPiwik::getPiwikUrl();
         if (! empty($newUrl)) {
--- a/SystemSettings.php
+++ b/SystemSettings.php
@@ -14,6 +14,9 @@ use Piwik\Plugins\TagManager\Context\BaseContext;
 use Piwik\Plugins\TagManager\Model\Environment;
 use Piwik\Settings\Setting;
 use Piwik\Settings\FieldConfig;
+use Piwik\Validators\NotEmpty;
+use Piwik\SettingsPiwik;
+use Piwik\Config;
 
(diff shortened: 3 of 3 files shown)


Template/Variable/MatomoConfigurationVariable.php
```diff
@@ -42,9 +42,10 @@ public function hasAdvancedSettings()
     public function getParameters()
     {
         $idSite = Common::getRequestVar('idSite', 0, 'int');
+        $settings = new Settings();
+        $url = $settings->getAnalyticsUrl();
 
-        $url = SettingsPiwik::getPiwikUrl();
-        if (SettingsPiwik::isHttpsForced()) {
+        if ($settings->isHttpsForced()) {
```
tsteur: Not quite understanding this change @lockland 

`$url` seems unused, and `$settings->isHttpsForced` is just calling `SettingsPiwik::isHttpsForced()` so could as well keep the old code and keep it directly.

tsteur:
@lockland As this feature is probably not needed by say 97% of our users, I wonder if we would be better off using a config setting for this. Either in `config.ini.php` or `config.php`.

Also then wondering if we should implement this rather in core in case other users want to use this for `piwik.js/matomo.js` as well. Could do it only for the tag manager for now though.

Like we can add something to `config.php` like `'publicContainerEndpoint' => 'https://www....'` and then get this value using `StaticContainer::get('publicContainerEndpoint')`. You would be able to adjust this in your `config/config.php` file.

or we use `plugins/TagManager/Configuration.php` to introduce a new config ini setting and you would be able to configure this using `/config/config.ini.php`. 

Ideally we wouldn't make the UI too much complex when only few users need it.

Template/Variable/MatomoConfigurationVariable.php
lockland: I didn't understand why you thought $url is unused whereas this variable is processed on if/else statement below.
I've created a method isHttpsForced on SystemSettings because I want to remove SettingsPiwik dependency since, in my opinion, all settings should be handled by SystemSettings. As I have an object on the same context I just used it.

lockland:
@tsteur 

I understood your point and in a first moment I didn't modify the interface, but my users don't have access to config files and because of it I prefer to create UI fields to set up these defaults values instead of just use a config file.
My ideia from the beginning was just modify it for tagmanager plugin so I couldn't figure out other use case, but I don't really know if this is the correct approuch

tsteur:
@lockland sorry for getting back to you so late again. I've reviewed it again and we'd definitely prefer it to have this configured as described in https://github.com/matomo-org/tag-manager/pull/192#issuecomment-544294645 as by far almost all our users wouldn't need this setting and it would only confuse all the other users. The beauty is though once it is in DI there could be a separate plugin that lets you configure this through the UI.

Eg In a custom plugin there could be these system settings you added. And in that plugin could be also a `plugins/YourPluginName/config/config.php` which then overwrites these settings like 

```php
return [
  'tagmanager_analytics_url' => function () {
      $settings = new \Piwik\Plugins\YourPluginName\SystemSettings();
     return $settings->getAnalyticsUrl();
  }
];
```

Is this something that could work for you? Or maybe you don't need the feature anymore and found a different solution? Sorry again.

lockland:
> @lockland sorry for getting back to you so late again. I've reviewed it again and we'd definitely prefer it to have this configured as described in [#192 (comment)](https://github.com/matomo-org/tag-manager/pull/192#issuecomment-544294645) as by far almost all our users wouldn't need this setting and it would only confuse all the other users. The beauty is though once it is in DI there could be a separate plugin that lets you configure this through the UI.
> 
> Eg In a custom plugin there could be these system settings you added. And in that plugin could be also a `plugins/YourPluginName/config/config.php` which then overwrites these settings like
> 
> ```
> return [
>   'tagmanager_analytics_url' => function () {
>       $settings = new \Piwik\Plugins\YourPluginName\SystemSettings();
>      return $settings->getAnalyticsUrl();
>   }
> ];
> ```
> 
> Is this something that could work for you? Or maybe you don't need the feature anymore and found a different solution? Sorry again.

@tsteur Firstly I'd like to say thank you for your time analysing my request and trying to give me a solution.
I'm till using that approach because my main problem is replacing matomo/piwik url on MatomoVariable created by TagManager plugin when a new site is added as well as on tagmanager container install code. So my question this moment is: Is it possible to replace the result of those calls below?

https://github.com/matomo-org/tag-manager/blob/1f674678b461ce562bd33b0f3cf57230ac879d8b/Context/WebContext.php#L221

https://github.com/matomo-org/tag-manager/blob/1f674678b461ce562bd33b0f3cf57230ac879d8b/Template/Variable/MatomoConfigurationVariable.php#L47

Just for let you know, my matomo infraestructure looks like this [diagram](https://pasteboard.co/JtukqZC.png)

Do you thing that snippet sent from you solve that problem of refering different urls for analytics, containers and matomo gui?

tsteur:
@lockland I think in this case you could look how it replaces the URL in https://github.com/lockland/plugin-TrackerJsCdnSync/blob/master/TrackerJsCdnSync.php#L21-L22 and https://github.com/lockland/plugin-TrackerJsCdnSync/blob/master/TrackerJsCdnSync.php#L48-L61 . 

it's also described a bit in https://github.com/matomo-org/matomo/issues/13462#issuecomment-644993731

lockland:
@tsteur 
That it is buddy!!
I've forked that project days ago because I intented to read that code and understand what was done, but those days have passed and I forgot day by day to do that.
I think that it solve all my problems and I'm so happy for no need to run a patch command anymore.
So now I think we can close this issue.

Thanks a lot.

lockland:
@tsteur 
I'm applying that solution we're talking about I a new question emerge in my mind.
Do you know what event can I use to modify the url setted on MatomoVariable after a new site has added?

I would like to subscribe on `API.TagManager.addContainerVariable.end` event and then to request the API to update that variable.

Do you think that approach is good enough or there is another one?

tsteur:
In this case you would probably be listening to `API.TagManager.addContainerVariable` and then add the parameters needed to configure the variable. In the method you can then modify the parameters used to add the variable like this (not tested but something in that direction)

```php

    public function onAddContainerVariable(&$finalParameters)
    {
        if ($finalParameters['type'] === 'MatomoConfigurationVariable') {
            $finalParameters['parameters']['jsEndpoint'] = 'tracker.js';
        }
        
    }
```

If that doesn't work there's also the event `API.TagManager.getContainerVariable`

lockland:
@tsteur 

It works perfectly, but I needed to do one modification.
```php
    public function onAddContainerVariable(&$finalParameters)
    {
        if ($finalParameters['type'] === 'MatomoConfiguration') {
            $finalParameters['parameters']['jsEndpoint'] = 'tracker.js';
        }
        
    }
```

Now I think that we can definitively close that PR

tsteur:
Great!

PR #199, opened 2019-10-30
fix: Tag::updateContainerTag setting a wrong startDate and endDate value

lockland:
When we do a tag update startDate and endDate is submitted as a empty
string that is converted to 0000-00-00 00:00:00 on database. That
conversion cause an ExceptionInvalidDateBeforeFirstWebsite exception
when We try to update tags again or publish a new version broken the
whole tagmanager plugin

### DIFF
--- a/Model/Tag.php
+++ b/Model/Tag.php
@@ -145,8 +145,8 @@ class Tag extends BaseModel
                 'fire_limit' => $fireLimit,
                 'fire_delay' => $fireDelay,
                 'priority' => $priority,
-                'start_date' => $startDate,
-                'end_date' => $endDate
+                'start_date' => empty($startDate) ? null : $startDate,
+                'end_date' => empty($endDate) ? null : $endDate,
             );
             $this->updateTagColumns($idSite, $idContainerVersion, $idTag, $columns);
         }


mikkoHonkanen:
We are facing the same bug as described in the original ticket. We already tried to fix it by updating the config with the provided solution but it doesn't solve the error. Creating and editing tags are OK but once creating a new version the Tag Manager breaks on error: 
"Start date: The date ‘-0001-11-30’ is a date before first website was online. Try date that’s after Aug 6, 1991 (timestamp 681436800).: -62169984000"

lockland:
> We are facing the same bug as described in the original ticket. We already tried to fix it by updating the config with the provided solution but it doesn't solve the error. Creating and editing tags are OK but once creating a new version the Tag Manager breaks on error:
> "Start date: The date ‘-0001-11-30’ is a date before first website was online. Try date that’s after Aug 6, 1991 (timestamp 681436800).: -62169984000"

Hello @mikkoHonkanen, to really solve that problem you have to update your database setting all invalid dates to null after have the patch applied. The query is below.

```
update matomo_tagmanager_tag set start_date =null, end_date = null where start_date = '0000-00-00 00:00:00';
```

mikkoHonkanen:
@lockland thank you! That fixed our problem,

PR #201, opened 2019-11-06
Added support for CD and CV  to Matomo Tag and Variables

MichaelRoosz:
- Added support for Custom Dimensions and Custom Variables into the Matomo Tag
- Added support for Custom Variables into the Matomo Variable

I think some tests still need to be adjusted, I will update the PR to fix them.

Requires https://github.com/matomo-org/matomo/pull/15125.

### DIFF
--- a/Template/Tag/MatomoTag.php
+++ b/Template/Tag/MatomoTag.php
@@ -137,7 +137,84 @@ class MatomoTag extends BaseTag
                     }
                     return $value;
                 };
-            })
+            }),
+            $this->makeSetting('customDimensions', array(), FieldConfig::TYPE_ARRAY, function (FieldConfig $field) {
+                $field->title = 'Custom Dimensions';
+                $field->description = 'Optionally set one or multiple custom dimensions.';
+                $field->validate = function ($value) {
+                    if (empty($value)) {
+                        return;
+                    }
+                    if (!is_array($value)) {
+                        throw new \Exception('Value needs to be an array');
+                    }
+                };
+
+                $field->transform = function ($value) {
+                    if (empty($value) || !is_array($value)) {
+                        return array();
+                    }
+                    $withValues = array();
+                    foreach ($value as $dim) {
+                        if (!empty($dim['index']) && !empty($dim['value'])) {
+                            $withValues[] = $dim;
+                        }
+                    }
+
+                    return $withValues;
+                };
+
+                $field->uiControl = FieldConfig::UI_CONTROL_MULTI_TUPLE;
+                $field1 = new FieldConfig\MultiPair('Index', 'index', FieldConfig::UI_CONTROL_TEXT);
+                $field1->customUiControlTemplateFile = self::FIELD_TEMPLATE_VARIABLE;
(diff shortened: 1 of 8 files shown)


Template/Tag/MatomoTag.php
```diff
@@ -137,7 +137,84 @@ public function getParameters()
                     }
                     return $value;
                 };
-            })
+            }),
+            $this->makeSetting('customDimensions', array(), FieldConfig::TYPE_ARRAY, function (FieldConfig $field) {
```
tsteur: @MichaelHeerklotz Not 100% sure I understand why they are also needed in the Tag and not just in the variable?

MichaelHeerklotz: As I understand it, triggering Custom Variables or Dimensions in a Tag gives more flexibility, and allows having multiple tags with different Custom Vars/Dims while using the same Matomo Config Variable.

tsteur: OK. In case you can find out just also for our understanding: Why is it better to have multiple tags over multiple configurations? Of course when you have multiple variables then it means you still need multiple tags so it be maybe a bit more work using the variables. Just wanting to avoid suddenly having heaps of configurations in the tag that are also in the variable (and more and more coming in their over time) and confusing users because there are two ways to do things. I can kind of see it though for custom dimensions and custom variables as they might be in PAGE/ACTION scope and the variable should maybe be used mostly for VISIT scope.

Wonder if we should explain this in the description that these are maybe mostly meant for custom dimensions in scope action/page?

MichaelRoosz: Yes, I think so, too. I think it makes sense having them both in the tag and the variable because it makes sense if they are of scope page.

Some explantion would help for sure.

MichaelRoosz: Should I add some explanations? Is there anything else missing for a merge?

tsteur: Nothing else missing to merge but the explanation. This will help users to know when to use what eg MatomoConfig custom dimensions vs Matomo tag custom dimensions. 

We'll need to fix the tests then but can do this after merging.

tsteur: @MichaelRoosz any chance you could add the explanation as discussed so we can merge?

peterbo: @tsteur @MichaelRoosz this is indeed an issue, that we're also running into on a regular base. 

An example:
We're tracking a page view with some content tags, custom page title, content groups, drilldown variables, etc. that only apply to the pageView. Any consecutive Events that are fired on that page also receive those custom dimensions, even if I fire the event as a "Custom HTML" tag. This makes the action custom dimensions that were collected with the pageview unusable (since they're mixed up with the actions that were triggered by the events).
I tried to create different configurations of every event, to be in better control of the custom dimensions that were then attached to the events, but that results in https://github.com/matomo-org/tag-manager/issues/242 (event is not attached to the corresponding pageView anymore). Apart from that, defining a complete configuration just for a "small" event is kind of overkill.

peterbo: @MichaelRoosz this would still be nice to have. Can I assist anywhere to get this done?

Issue #214, opened 2020-01-11
Report Forms variable typo


Laul0:
- Plateform: self-hosted (docker)
- Matomo version: 3.13.0

## Current behavior

From the Tag Manager variables, in the Forms area, it is possible to retrieve the "Form Destination URL" and the variable name is `{{FormDestionation}}`

![Screen Shot 2020-01-07 at 09 48 30](https://user-images.githubusercontent.com/15279205/72198015-1bd2e980-33f6-11ea-9891-dbe088d1d944.png)

## Expected behavior

The variable name should be `{{FormDestination}}`

tsteur:
Cheers @Laul0 
we'll change this as part of Matomo 4 in case the variable was already used previously because it could potentially break someones container if we just renamed the variable.

PR #215, opened 2020-01-12
Fix form destination name

tsteur:
fix https://github.com/matomo-org/tag-manager/issues/214


diosmosis:
There are still some test failures that seem off (after changing the base branch)

tsteur:
Thanks @diosmosis  fixed

PR #220, opened 2020-02-03
Use renamed CustomJsTracker plugin

sgiehl:
refs matomo-org/matomo#15505


tsteur:
Looks generally good, just need to wait re clarifying plugin name in  matomo-org/matomo#15505

tsteur (approved):

PR #222, opened 2020-02-06
Make sure clicks are tracked even if event propagation is stopped

### LINKED ISSUE #221: Unable to attach event listener to element if theres already an event listener that stops bubbling
Per definition: If an element has multiple event handlers on a single event, then even if one of them stops the bubbling, the other ones still execute.

But that doesn't seem to be the case. Is this caused by a certain implementation detail or could this be a non expected behavior?

Example:
`<a href="#" class="new-message-button text-uppercase"><span class="short">Nachricht</span><span class="long">Ihre Nachricht an uns</span></a>`

The a- element has an event-listener that stops bubbling: 
`function() {
  return $(this).parents(".new-message-box").addClass("open"), !1
}`

Attaching an "All Links Click" Trigger has no effect. As soon as the first event-listener is deleted, the "All Links Click" trigger starts to fire onclick.

tsteur:
fix https://github.com/matomo-org/tag-manager/issues/221

Haven't tested it yet if it actually fixes the issue. @peterbo created this PR maybe they can give it a try?


tsteur:
Reading https://gomakethings.com/when-to-use-use-capture-in-your-event-listeners/ and https://www.w3.org/TR/2003/NOTE-DOM-Level-3-Events-20031107/events.html#Events-phases

Not sure we want to merge it. To be tested fully. I reckon it might just work but not sure.

peterbo:
Tested the changes extensively in two different, quite complex TagManager environments. No unexpected behaviour so far. Still searching for unexpected edge-cases.

tsteur:
@peterbo I did some testing and debugging as well and it works for me too and can't think of any possible issues.

peterbo:
I'll be keeping an eye on weird edge cases. Webtrends is using it also (besides GTM). I think this looks good!

peterbo:
Hey @tsteur - any chance to merge this also in the 3.x branch? This would be a highly appreciated fix for the LTS versions.

tsteur:
feel free to create a PR for it but don't know when the next release will be. BTW Matomo 4 will be released next week if this helps

PR #231, opened 2020-03-07
Unsanitize tag name in preview mode

### LINKED ISSUE #230: Tag names are not encoded
Hi there,
When I create a tag in Tag Manager with a ' within the title, the debug mode carries the name like &amp.
Could you please correct it?

tsteur:
fix https://github.com/matomo-org/tag-manager/issues/230

The tag/trigger/variable names are stored sanitized so for them to appear correctly in the preview they need to be unsanitzed. Angular should take care of the escaping in `debug.twig` template.


diosmosis (approved):
LGTM. Note: you can add xss testing easily by using `XssTesting` to generate xss content for names/etc. The UI tests will fail if there's a bug and one triggers an issue.

PR #238, opened 2020-04-14
etracker Configuration update - cookieless tracking by default

ben-dw:
Updated the configurtaion file for cookieless tracking (now: by default).
Furthermore checked if the etracker function is already loaded before an event is fired (etracker sometimes loads too slow).


tsteur:
Thanks a lot @ben-dw

PR #240, opened 2020-04-22
etracker wrapper function added for single page applications compatibility

ben-dw:
I have added one further function of etracker (https://www.etracker.com/docs/integration-setup/tracking-code-sdks/tracking-code-integration/dynamische-inhalte-wrapper/).
This function is needed for SPA (single page applications) to trigger pageviews.

### DIFF
--- a/Template/Tag/EtrackerTag.php
+++ b/Template/Tag/EtrackerTag.php
@@ -28,6 +28,7 @@ class EtrackerTag extends BaseTag
             $field->validators[] = new NotEmpty();
             $field->availableValues = array(
                 'pageview' => 'Pageview',
+                'wrapper' => 'Wrapper',
                 'event' => 'Event',
             );
         });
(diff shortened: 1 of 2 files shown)


tsteur:
Cheers @ben-dw

ben-dw:
I'm wondering why the changelog 3.13.5 shows that this ticket is integrated. I can't see any changes in the download file for the new matomo version.
@tsteur : Do you have any idea?

sgiehl:
That actually shouldn't be included in the changelog. It was merged to 4.x-dev branch, so it will be included in the first Matomo 4 release and not in any 3.x release. ping @mattab

Findus23:
I think the milestone would need to be ignored here: 
https://github.com/matomo-org/github-changelog-generator/blob/a85da2b0184740ee010f74cf131f4334f9182b5b/src/config.js#L28-L32

mattab:
now deleted the entries from changelog, thanks for the notice @ben-dw 

(also renamed the milestone to Matomo 4.0.0 so it would match my regex I had set: `4.0*`)

PR #244, opened 2020-05-18
better CSS Selector help

Findus23:
reported in https://forum.matomo.org/t/element-visibility-trigger-doesnt-work-for-css-class/35324

while this might be obvious to everyone who knows CSS, there are probably many Tag Manager user who don't know CSS at all.


tsteur:
Thanks @Findus23 👍

PR #252, opened 2020-06-17
Improve detection of preview release if it wasn't detected previously somehow

tsteur:
…


mattab:
@tsteur Customer mentioned that this has helped a bit. could we merge this?

tsteur:
"helped a bit" (I reckon there the problem was actually the file permissions?) I doubt it actually made any change but will merge anyway just in case as it doesn't hurt to have it.

PR #258, opened 2020-07-08
Adjust code to update window._mtm

tsteur:
refs https://github.com/matomo-org/matomo/issues/16189
refs https://github.com/matomo-org/matomo/issues/16055

@Findus23 can you have a look here too? Will then merge and update submodule in 3.x

recreated PR in https://github.com/matomo-org/tag-manager/pull/259


tsteur:
closing as I think the wrong base was used. Will recreate PR

PR #259, opened 2020-07-08
Adjust code to update window._mtm

tsteur:
refs https://github.com/matomo-org/tag-manager/pull/258
refs https://github.com/matomo-org/matomo/issues/16189
refs https://github.com/matomo-org/matomo/issues/16055

@Findus23 can you have a look here too? Will then merge and update submodule in 3.x


Findus23 (approved):
The change should be fine. But I would mention it in the changelog to not confuse users if their old tracking code is "broken" and needs to be replaced.

PR #260, opened 2020-07-09
 use Array for dataLayer

### LINKED ISSUE #257: TagManager _mtm dataLayer is an object not an array -> use Array for dataLayer
In contrast to GA's dataLayer array, Matomo's _mtm datalayer is an object. Quite a lot of solutions, when it comes to consent management, plugins for CMS, shop integration plugins and so forth, implement a method to rename the datalayer variable name nowadays (to use another solution than GA). In most cases, this fails with _mtm, because it is not an array (it just implements some array methods).

I was wondering, what are the advantages/disadvantages of _mtm being an object vs being an array. Making _mtm an array could make Matomo's datalayer more compatible to a lot of tools out there.

tsteur:
fix #257


peterbo:
Hi @tsteur - thanks for the ongoing changes! With the last change, I get some kind of circular reference in the _mtm array:

circular-reference

tsteur:
Sorry about that @peterbo I now changed it and actually tested it and it should now work 👍

PR #283, opened 2020-10-27
Offer possibility to only initialise the tracker without tracking any data

### LINKED ISSUE #282: When using _paq.push(['trackPageView']); in a Custom HTML tag (with bundled and default tracker in configuration), no tracking request is fired
When no "Matomo Analytics" prepared pageView tag is fired, the _paq.push call will not trigger a pageView request and _paq stays an array. The tracker object doesn't seem to be initialised without firing a "Matomo Analytics" prepared pageView tag.

tsteur:
fix #282 

fyi @peterbo

### DIFF
--- a/Template/Tag/MatomoTag.php
+++ b/Template/Tag/MatomoTag.php
@@ -45,6 +45,7 @@ class MatomoTag extends BaseTag
                 'pageview' => 'Pageview',
                 'event' => 'Event',
                 'goal' => 'Goal',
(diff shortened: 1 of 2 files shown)


peterbo:
Hi @tsteur, thank you! I tested the change in connection with Custom HTML tags and it fixes the problem. The tracker gets initialised and the Custom HTML pageView is triggered correctly. Also in connection with the Matomo Analytics prepared tags, everything works well.

PR #285, opened 2020-11-05
remove stack trace from error message

diosmosis:


TagManager.php
```diff
@@ -279,7 +280,9 @@ public function regenerateReleasedContainers()
                     }
                 }
             } catch (\Exception $e) {
-                Log::error('There was an error while regenerating container releases: ' . $e->getMessage()  .$e->getTraceAsString());
+                StaticContainer::get(LoggerInterface::class)->error('There was an error while regenerating container releases: {ex}', [
```
tsteur: @diosmosis here as well could we add the trace if the trace is enabled?

diosmosis: @tsteur should be done automatically if ExceptionToTextProcessor is enabled

diosmosis: Oh, the var name is wrong

tsteur: 👍  OK LGTM then. Be great to double check though it works as expected

diosmosis: Verified that when the following is in bootstrap.php:

```
define('PIWIK_PRINT_ERROR_BACKTRACE', 1);
```

the stacktrace chain will appear in log output, when, for instance, plugins are deactivated/activated through the UI. This does not happen in CLI mode, however, since Monolog/cli.php overwrites the log handlers.

PR #287, opened 2020-11-24
Add consent methods to Matomo Tag Manager

tsteur:
needs to be tested


diosmosis:
Tested locally, works for me. Not sure if there should be automated tests, but will merge.

PR #289, opened 2020-12-07
Fix return value of usort

sgiehl:
Returning other values than integer triggers a deprecation on PHP 8

Running tests on PHP 8 triggers this a lot:

```
Deprecated: usort(): Returning bool from comparison function is deprecated, return an integer less than, equal to, or greater than zero in /home/travis/build/matomo-org/matomo/plugins/TagManager/Dao/TagsDao.php on line 210
```

### DIFF
--- a/Dao/TagsDao.php
+++ b/Dao/TagsDao.php
@@ -206,7 +206,7 @@ class TagsDao extends BaseDao implements TagManagerDao
                 $indexB = array_search($tagB, $tags);
                 return $indexA - $indexB;
             }
-            return $tagA['priority'] > $tagB['priority'];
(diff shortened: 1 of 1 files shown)


tsteur:
LGTM Can you check this part is covered by a test to make sure it works @sgiehl ?

sgiehl:
@tsteur had the parameter order reversed before, and quite a lot tests failed, so guess it should be tested. See https://travis-ci.com/github/matomo-org/tag-manager/builds/207027632

tsteur:
sweet. feel free to merge (will need to update submodule then)

PR #295, opened 2021-01-18
Prevent possible circular reference in preview mode in addDebugValues

nova-os:
Prevent "TypeError: Converting circular structure to JSON" similar to https://github.com/matomo-org/tag-manager/pull/186fpr html elements in the data layer.

React.js does set circular references to the dom elements, which lead JSON.stringify to fail them:

Bildschirmfoto 2021-01-18 um 08 29 40


tsteur:
Thanks @nova-os very much appreciated

PR #307, opened 2021-03-22
force run onSiteChange first time

flamisz:
refs 
matomo-org/matomo#16787
matomo-org/matomo#17367

### DIFF
--- a/angularjs/tagmanagerTrackingCode/tagmanager.controller.js
+++ b/angularjs/tagmanagerTrackingCode/tagmanager.controller.js
@@ -26,6 +26,7 @@
         this.releases = [];
         this.installInstructions = [];
         this.noReleaseFound = false;
+        this.firstTime = true;
         var translate = $filter('translate');
         var ucfirst = $filter('ucfirst');
 
@@ -42,6 +43,7 @@
             this.environments = [];
             this.matomoConfigs = [];
             this.idContainer = '';
+            this.firstTime = false;
 
             if (!this.site || !this.site.id) {
                 return;
@@ -187,7 +189,10 @@
             if (val !== oldVal) {
                 self.onSiteChange();
             }
-        });
 
+            if (val === oldVal && self.firstTime) {
+                self.onSiteChange();
+            }
+        });
     }
 })();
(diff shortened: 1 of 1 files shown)


flamisz:
@tsteur, @diosmosis already merged [this PR](https://github.com/matomo-org/matomo/pull/17367). 
This is connected to that one, because the tag manager containers are not loading onto the page without this. Can you please you or @sgiehl check it?

angularjs/tagmanagerTrackingCode/tagmanager.controller.js
```diff
@@ -187,7 +189,10 @@
             if (val !== oldVal) {
                 self.onSiteChange();
             }
-        });
 
+            if (val === oldVal && self.firstTime) {
+                self.onSiteChange();
+            }
+        });
```
diosmosis, line 192: Do you think it would be better to combine the two `if`s into `if (val !== oldVal || self.firstTime)`?

diosmosis:
I'm not super familiar w/ the tag manager, but the code looks good to me

angularjs/tagmanagerTrackingCode/tagmanager.controller.js
flamisz: you're right, that would be the same, but simpler. I just added this new check separately. I'll combine those.

sgiehl:
@flamisz  could you maybe describe a bit clearer what "bug" this should fix?

flamisz:
> @flamisz could you maybe describe a bit clearer what "bug" this should fix?

@sgiehl sure.
For the PR I needed this change, I included the `tagmanager` component. The same, what we use on the *admin/tracking code* page. 
It works there, but not working here, because it's not calling that `onSiteChange` when the page loads. I `console.log`-d the `val` and `oldVal` in the `$watch` method, and this is how it looks like in the admin page and on the "no data yet page":
```
// admin page
undefined undefined
8 undefined

// no data yet page
8 8 
```

I'm not sure where that `undefined` is coming from but looks like it almost just working accidentally. So I added this "firstTime" check.
Now that I'm thinking through, maybe a better way, like run that when the component loads, but this is working too.

sgiehl (approved):
Now I was able to reproduce and test. Seems to work as expected now.

PR #317, opened 2021-04-08
Validate tag manager variable name

### LINKED ISSUE #308: Using special characters in variable name results in errors in trigger configuration
Currently it seems when creating a variable it's possible to use a `/` or `&` in the variable name. Doing so however results in the following error when creating a trigger using that variable (For example): 
`Conditions: The variable "URL Variable with / and &" in the condition at position "0" cannot be found.`

image

image

We should either prevent users from being able to add special characters to variable names or sanitize the variable name to prevent issues like this.

flamisz:
fixes: #308 

With this validation, we only let alphanumeric characters and spaces in a variable name. There is another PR for this at matomo repository because I added both front and back en validation, and a new language variable.

But after this change, when someone wants to update a variable, that already worked but contained special characters, can not make the change without renaming it. Are we ok with that?

### DIFF
--- a/angularjs/manageVariable/edit.controller.js
+++ b/angularjs/manageVariable/edit.controller.js
@@ -258,10 +258,32 @@
             return true;
         }
 
+        function validateNameField() {
+            var regex = /^[a-zA-Z0-9\s]+$/;
+            var title = _pk_translate('General_Name');
+
+            if (!self.variable.name) {
+                return true;
+            }
+
+            if (!regex.test(self.variable.name)) {
+                showNotification(_pk_translate('General_ValidatorErrorNoValidAlphaNumSpaces', [title]), 'error');
+
+                return false;
+            }
+
+            return true;
+        }
+
+
         this.createVariable = function () {
             removeAnyVariableNotification();
 
             if (!checkRequiredFieldsAreSet()) {
+                return;
+            }
+
+            if (!validateNameField()){
                 return;
             }
 
@@ -312,6 +334,10 @@
             removeAnyVariableNotification();
 
             if (!checkRequiredFieldsAreSet()) {
+                return;
+            }
+
+            if (!validateNameField()){
                 return;
             }
 
--- a/angularjs/manageVariable/edit.controller.js
+++ b/angularjs/manageVariable/edit.controller.js
@@ -317,9 +317,7 @@
                 return;
             }
 
-            if (!validateNameField()){
-                return;
-            }
+            this.variable.name = encodeURIComponent(this.variable.name);
 
             this.isUpdating = true;
 
--- a/angularjs/manageVariable/edit.controller.js
+++ b/angularjs/manageVariable/edit.controller.js
@@ -258,24 +258,6 @@
             return true;
         }
 
-        function validateNameField() {
-            var regex = /^[\p{L}\p{M}\p{N} ]+$/u;
-            var title = _pk_translate('General_Name');
-
-            if (!self.variable.name) {
-                return true;
-            }
-
-            if (!regex.test(self.variable.name)) {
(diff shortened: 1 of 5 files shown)


flamisz:
The new validator PR is here: matomo-org/matomo#17435

angularjs/manageVariable/edit.controller.js
```diff
@@ -258,13 +258,35 @@
             return true;
         }
 
+        function validateNameField() {
+            var regex = /^[a-zA-Z0-9\s]+$/;
```
sgiehl: That actually does not do the same as the validator on PHP side, is that on purpose? It won't allow foreign languages and also white space characters that are not a simple space.

flamisz: yeah I know, I wanted to look into this but forgot. The PHP version is much better, I will "translate" that into JS.

flamisz: Hi @sgiehl,
I can't use the same, unicode version of the regex as in PHP, because IE does not support it.
Should we leave the front-end validation? Or should I try to write a regex without unicode?

We still support IE11, don't we?

flamisz: FYI, I tried [babel](https://babeljs.io/repl#?browsers=defaults%2C%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&spec=false&loose=false&code_lz=DYUwLgBATiDmIA8IF4IHoB6BtAOgBwG8AZAX3wIFkzCA5EiAXQGoASNAVwChRIBnMKCggByAIyjhnTjHgIAdGBD8AFPygBKKZyA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2&prettier=false&targets=&version=7.13.15&externalPlugins=), and got a very long regex, and still not sure it supports every language.

I would probably leave the front-end validation as I can't hide that line from IE. Or is there a way to "hide" it?

sgiehl: In that case it might be better not to validate here at all and let the backend return the error message if it doesn't match.
Otherwise we should adjust the validation in backend to match the one in javascript.
Not sure if foreign languages should be allowed in names at all. ping @tsteur

tsteur:
as discussed we don't need to validate in JS and be good to check for the root cause if maybe there is just an issue with conditions in the trigger.

angularjs/manageVariable/edit.controller.js
```diff
@@ -265,6 +265,8 @@
                 return;
             }
 
+            this.variable.name = encodeURIComponent(this.variable.name);
```
sgiehl: That works, but makes the UI showing the encoded value while the request is running. Might be better to make a local copy of the variable and only manipulate the value there to avoid this

flamisz: yes, true. I'll add that.

angularjs/manageVariable/edit.controller.js
```diff
@@ -267,7 +267,10 @@
 
             this.isUpdating = true;
 
-            tagManagerVariableModel.createOrUpdateVariable(this.variable, 'TagManager.addContainerVariable').then(function (response) {
+            var tempVariable = JSON.parse(JSON.stringify(this.variable));
```
sgiehl: instead of JSON parse/stringify you could also simply use `angular.copy`

flamisz: thanks, didn't know about that. Done

sgiehl (approved):

PR #327, opened 2021-05-24
Use SERVER_PORT if available.

q464q:
謝謝


sgiehl (approved):
Haven't tested, but code looks good.

Issue #328, opened 2021-05-26
Observe DOM element changes even when a user doesn't scroll on the page


mattab:
currently in the "Element visibility" trigger (and possibly others), the DOM is re-scanned (only?) when a "Scroll" is detected. But sometimes, a new Element might become visible, yet the user is not scrolling (for example when the user clicks on a button and a new DOM element appears). The Element visibility trigger wouldn't be triggered in this case (until the user scrolls). 

In terms of solutions, I have no idea really.... but some thoughts anyway:
maybe there could be some possible improvements (ie. detect element visibility on "Click" as well as "Scroll" and maybe other events also).

heurteph-ei:
I think if you solve #B, you'll solve #A.
But could "Observe DOM change" feature detect for example a pseudo-element displayed by CSS on the :hover selector?

ishantsolanki-vakt:
I am facing exactly the same issue.
I initially thought the `Element Visibility Trigger` would serve the purpose. But i was wrong. It only works (in my case) when i resize the window.
I would like to capture events like When a modal was opened or when showing an error message for an incorrect validation. In such cases there is no scroll events happening.

tsteur:
fyi splitted this into two issues: https://github.com/matomo-org/tag-manager/issues/339 because this one is a lot easier to implement compared to #339 and can therefore be put in current sprint etc

Issue #339, opened 2021-08-03
Visibility trigger add feature to observe DOM changes etc.


tsteur:
from https://github.com/matomo-org/tag-manager/issues/328

we got another report from a user, which reads:

> On our registration page, we have a few steps that we want to track (email, details, password, [optionally] verification). They are each assigned an unique element ID, which we use in the tag manager, but only the first step is seen as visible. If we resize the window the visibility of the current step is triggered, because we do a layout change. But just clicking through the steps do not trigger the visibility triggers, despite the ID of the step existing in the DOM and being visible. **Google Tag Manager's visibility triggers works just fine for these steps, where there is an option to enable "Observe DOM changes". An option seemingly missing in Matomo Tag Manager.**

Here is what the element visibility trigger looks like in GTM: 

![image](https://user-images.githubusercontent.com/466765/119725870-948bf600-bec4-11eb-9d1d-5d57375b92bb.png)

Maybe we could introduce a feature "Observe DOM changes" like in GTM?

justinvelluppillai:
@AltamashShaikh is this issue fixed and can be closed now?

Issue #329, opened 2021-05-27
Users with "view" permissions have access to the Tag Manager


Jaimeapo:
Users with "view" permissions have access to the Tag Manager, something that was not possible in previous versions 3.13.6.
Is there any intention of this to be the case as of version 4.1.1.?

Fortunately, it is not possible to make changes, but an error message appears "Your session has expired due to inactivity. Please log in to continue." But it goes back, in the browser, you can continue browsing ... (your session has not expired actually)
Which is also inconsistent, as something like "you are not allowed to make changes" should appear.

diosmosis:
Hi @Jaimeapo, this appears to be intentional and was added in https://github.com/matomo-org/tag-manager/pull/73. Users w/ view permissions should only be able to view the information, they should not even be able to see icons required to go into edit mode. Is this not the case for you? If not, can you provide some screenshots to show us what you see (for example, the manage containers page w/ the icons accessible and the error message you see and where it occurs)?

tsteur:
@Jaimeapo as @diosmosis it is indeed intentional that the configurations can be viewed but not edited. You should however not see any message like "Your session has expired". I've just tried to reproduce this using the latest version of Matomo and couldn't. I'll close this issue for now but if you still have this issue please comment with steps to reproduce this and we'll reopen again and have a look into it. Thanks for this.

PR #341, opened 2021-08-19
Implemented Intersection Observer to fire visibility trigger along with scroll and DOM load

AltamashShaikh:
Implemented Intersection Observer to fire visibility trigger along with scroll and DOM load

### Description:

The code will now also use Intersection Observer to detect visibility
This will help us to track elements which gets visible when there is no scroll event.
It will automatically stop observing if trigger once per page or once per element is set.
This will fix the issue #328 

### Review

### DIFF
--- a/Template/Trigger/ElementVisibilityTrigger.web.js
+++ b/Template/Trigger/ElementVisibilityTrigger.web.js
@@ -243,22 +243,24 @@
         }
 
         function setIntersectionObserver(triggerEvent) {
-            if ('IntersectionObserver' in window) {
-                var interSectionObserverOptions = {
-                    root: null, // document's viewport as the container.
-                    rootMargin: '0px',
-                    threshold: (minPercentVisible / 100)
-                };
-                observerIntersection = new IntersectionObserver(function (entries) {
-                    interSectionCallback(entries, triggerEvent);
-                }, interSectionObserverOptions);
+            return function () {
+                if ('IntersectionObserver' in window) {
+                    var interSectionObserverOptions = {
+                        root: null, // document's viewport as the container.
+                        rootMargin: '0px',
+                        threshold: (minPercentVisible / 100)
+                    };
+                    observerIntersection = new IntersectionObserver(function (entries) {
+                        interSectionCallback(entries, triggerEvent);
+                    }, interSectionObserverOptions);
 
-                if (selectors) {
-                    TagManager.dom.bySelector(selectors).forEach(function (element) {
-                        observerIntersection.observe(element);
-                    });
+                    if (selectors) {
+                        TagManager.dom.bySelector(selectors).forEach(function (element) {
+                            observerIntersection.observe(element);
+                        });
+                    }
                 }
-            }
+            };
 
         }
 
--- a/Template/Trigger/ElementVisibilityTrigger.web.js
+++ b/Template/Trigger/ElementVisibilityTrigger.web.js
@@ -11,6 +11,7 @@
         var blockTrigger = false;
         var onlyOncePerElement = fireTriggerWhen === 'onceElement';
         var selectors = getSelectors();
+        var observerIntersection;
 
         function getPercentVisible(node)
         {
@@ -189,16 +190,10 @@
 
                 var dom = TagManager.dom;
 
-                if (selectionMethod === 'elementId') {
-                    var node = dom.byId(parameters.get('elementId'));
-                    if (node) {
-                        nodes.push(node);
-                    }
-                } else if (selectionMethod === 'cssSelector') {
-                    nodes = dom.bySelector(parameters.get('cssSelector'));
-                } else {
+                if (!selectors) {
                     return;
                 }
+                nodes = TagManager.dom.bySelector(selectors);
 
                 for (var i = 0; i < nodes.length; i++) {
                     if (onlyOncePerElement) {
@@ -221,8 +216,14 @@
                             if (fireTriggerWhen === 'oncePage') {
                                 blockTrigger = true;
                                 TagManager.window.offScroll(self.scrollIndex);
+                                if (observerIntersection) {
+                                    observerIntersection.disconnect();
+                                }
                             } else if (onlyOncePerElement) {
                                 triggeredNodes.push(nodes[i]); // to avoid possible memory leaks as much as possible we add onceElement only when needed
+                                if (observerIntersection) {
+                                    observerIntersection.unobserve(nodes[i]);
+                                }
                             }
                         }
                     }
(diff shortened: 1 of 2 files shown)


tsteur:
@AltamashShaikh could we add an option to the element visibility trigger named ok "Observe DOM changes" and it would be disabled by default?  Similar to the image in https://github.com/matomo-org/tag-manager/issues/339

image

You can add such an option here: https://github.com/matomo-org/tag-manager/blob/4.x-dev/Template/Trigger/ElementVisibilityTrigger.php#L28

We would then use scroll, dom.onLoad and click event listener to check for visibility changes in case some elements are added later. We could maybe still use the intersection part for this but be still good to detect these changes.

When the observe feature is enabled, we would listen to the mutation observer.

tsteur:
Also could we stop checking for the selector if the trigger is configured to be triggered only once similar to `                                blockTrigger = true;TagManager.window.offScroll(self.scrollIndex);` in https://github.com/matomo-org/tag-manager/pull/341/files#diff-910727e972309dfa19d61e5804fcea43485b3fb03556e7419dfe842479b7180aL228 ? 

Haven't otherwise looked much yet at the other logic.

AltamashShaikh:
> Also could we stop checking for the selector if the trigger is configured to be triggered only once similar to ` blockTrigger = true;TagManager.window.offScroll(self.scrollIndex);` in https://github.com/matomo-org/tag-manager/pull/341/files#diff-910727e972309dfa19d61e5804fcea43485b3fb03556e7419dfe842479b7180aL228 ?
> 
> Haven't otherwise looked much yet at the other logic.

Yes we are executing observer.unobserve for such cases.
Refer: https://github.com/matomo-org/tag-manager/blob/1f9a3c18d398971f15234d9b92f7ddcef414c69e/Template/Trigger/ElementVisibilityTrigger.web.js#L284

AltamashShaikh:
> @AltamashShaikh could we add an option to the element visibility trigger named ok "Observe DOM changes" and it would be disabled by default? Similar to the image in #339
> 
> image
> 
> You can add such an option here: https://github.com/matomo-org/tag-manager/blob/4.x-dev/Template/Trigger/ElementVisibilityTrigger.php#L28
> 
> We would then use scroll, dom.onLoad and click event listener to check for visibility changes in case some elements are added later. We could maybe still use the intersection part for this but be still good to detect these changes.
> 
> When the observe feature is enabled, we would listen to the mutation observer.

We wont be adding this change in the PR.
We will add the observer DOM change as an option in the PR for #329

Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -235,9 +230,82 @@
             };
         }
 
+        function getSelectors() {
```
tsteur, line 234: fyi in line 122-197 maybe we can now as well use `TagManager.dom.bySelector(selectors).`?`

Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -235,9 +230,82 @@
...
+        function interSectionCallback(entries, triggerEvent, observer) {
+            var dom = TagManager.dom;
+            entries.forEach(function (entry) {
+                if (entry.intersectionRatio > 0) {
+                    if (blockTrigger || (onlyOncePerElement && isNodeEventTriggered(entry.target))) {
+                        return;
+                    }
+                    triggerEvent({
+                        event: 'mtm.ElementVisibility',
+                        'mtm.elementVisibilityPercentage': minPercentVisible,
```
tsteur: Is there a way to get the actual percentage with intersection observer? is it maybe `intersectionRatio`? Be great to get the actual value there

tsteur: otherwise we could fallback to `getPercentVisible(node)` maybe

Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -235,9 +230,82 @@
...
+                }
+            }
+
+            return false;
+        }
+
         this.setUp = function (triggerEvent) {
             this.scrollIndex = TagManager.window.onScroll(checkVisiblity(triggerEvent));
             TagManager.dom.onLoad(checkVisiblity(triggerEvent));
+            setIntersectionObserver(triggerEvent);
```
tsteur: I'm thinking we shouldn't load the intersection observer at this point. If the page is not fully loaded yet, then some elements may be visible for just a tiny short time before the CSS is applied and the element pushed down maybe.

Maybe we could instead try to run this on "onLoad" like `TagManager.dom.onLoad(setIntersectionObserver(triggerEvent));`?

would intersection observer let us know if it was visible previously before onload? (just in case the onload event is quite delayed)

Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -235,9 +230,82 @@
...
+                        'mtm.elementVisibilityClasses': dom.getElementClassNames(entry.target),
+                        'mtm.elementVisibilityText': TagManager.utils.trim(entry.target.innerText),
+                        'mtm.elementVisibilityNodeName': entry.target.nodeName,
+                        'mtm.elementVisibilityUrl': entry.target.href || dom.getElementAttribute(entry.target, 'href'),
+                    });
+
+                    if (fireTriggerWhen === 'oncePage') {
+                        blockTrigger = true;
+                        TagManager.dom.bySelector(selectors).forEach(function (element) {
+                            observer.unobserve(element);
```
tsteur: would we check here and below if there is maybe `self.scrollIndex` defined and in this case also call `
                                TagManager.window.offScroll(self.scrollIndex);`?

tsteur: and where we call `TagManager.window.offScroll(self.scrollIndex);` would we maybe call `observer.unobserve(element);`?

Template/Trigger/ElementVisibilityTrigger.web.js
tsteur: or I think `checkVisiblity` detects elements that may have been visible before "onload" by checking the scrolling position and assumes any element above this would have been visible at some stage. Not sure.

In that case could also execute something like 

```
TagManager.dom.onLoad(checkVisiblity(triggerEvent)); // detect any elements that have been visible previously
TagManager.dom.onLoad(setIntersectionObserver(triggerEvent)); // detect any matching element when it becomes visible
```

tsteur (changes requested):
@AltamashShaikh 👍  looks already quite good. I haven't tested it but left a few comments to think about.

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: Refactored the code

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: Moved to dom onload as suggested above

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: Started using getPercentVisible along with max of minPercentVisible as discussed

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: Yes added both at the respective callbacks

tsteur:
Tested this for quite a while and worked nicely 👍

PR #344, opened 2021-08-27
Visibility trigger add feature to observe DOM changes

AltamashShaikh:
### Description:

Added code to observe dom changes via MutationObserver
Issue: #339 

### Review


Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -310,10 +294,86 @@
...
+                'mtm.elementVisibilityText': TagManager.utils.trim(node.innerText),
+                'mtm.elementVisibilityNodeName': node.nodeName,
+                'mtm.elementVisibilityUrl': node.href || dom.getElementAttribute(node, 'href')
+            });
+        }
+
         this.setUp = function (triggerEvent) {
-            this.scrollIndex = TagManager.window.onScroll(checkVisiblity(triggerEvent));
-            TagManager.dom.onLoad(checkVisiblity(triggerEvent));
+            if (!(isMutationObserverSupported && observeDomChanges && isIntersectionObserverSupported)) {
```
tsteur: @AltamashShaikh I think the code would be a more readable when writing eg 

```js
var useMutationObserver = isMutationObserverSupported && observeDomChanges && isIntersectionObserverSupported;
if (useMutationObserver) {
    TagManager.dom.onLoad(setMutationObserver(triggerEvent));
} else {
  this.scrollIndex = TagManager.window.onScroll(checkVisiblity(triggerEvent));
                TagManager.dom.onLoad(checkVisiblity(triggerEvent));
}
TagManager.dom.onLoad(setIntersectionObserver(triggerEvent));
```

When reviewing and looking at code it looked initially like mutation observer is always added etc.

lang/en.json
```diff
@@ -553,6 +553,8 @@
         "WindowUnloadTriggerName": "Window Unload",
         "ZendeskChatTagDescription": "Reach your customers, instantly via web, mobile, and messaging.",
         "ZendeskChatTagHelp": "This tag allows you to add the Zendesk Chat form to your website.",
-        "ZendeskChatTagName": "Zendesk Chat (formerly Zopim)"
+        "ZendeskChatTagName": "Zendesk Chat (formerly Zopim)",
+        "SettingElementVisibilityObserveDomChangesTitle": "Observe DOM Changes",
+        "SettingElementVisibilityObserveDomChangesDescription": "Enable if you want to track dynamic DOM elements."
```
tsteur: Could we mention that on few sites this may slow down the website if there are a lot of DOM changes?

Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -310,10 +294,86 @@
             return false;
         }
 
+        function setMutationObserver(triggerEvent) {
+            return function () {
+                if (observeDomChanges && isMutationObserverSupported) {
+                    var targetNode = documentAlias.body;
+                    var config = {attributes: true, childList: true, subtree: true};
+                    observerMutation = new MutationObserver(function (mutationsList) {
+                        mutationObserverCallback(mutationsList, triggerEvent);
```
tsteur: @AltamashShaikh wonder if we could apply similar behaviour as in the scroll listener in https://github.com/matomo-org/tag-manager/blob/4.x-dev/javascripts/tagmanager.js#L609-L617

to only trigger this event like every 120ms so if this event is triggered eg multiple times per second or even like every ms then it won't slow down the site as much?

Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -310,10 +294,86 @@
...
+                    });
+                    observerMutation.observe(targetNode, config);
+                }
+            };
+        }
+
+        function mutationObserverCallback(mutationsList, triggerEvent) {
+            for (var index in mutationsList) {
+                var mutation = mutationsList[index];
+                if (mutation.type === 'childList' || mutation.type === 'attributes') {
```
tsteur: just curious to we still need this check as we're only listening to `{attributes: true, childList: true, subtree: true};`?

Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -310,10 +294,86 @@
...
+                    var addedNodes = mutation.addedNodes;
+                    addedNodes.forEach(function (node) {
+                        TagManager.dom.bySelector(selectors).forEach(function (element) {
+                            if (node.contains(element)) {
+                                if (blockTrigger || (onlyOncePerElement && isNodeEventTriggered(element))) {
+                                    return;
+                                }
+
+                                if (!isNodeInViewport(element) && observerIntersection) {
+                                    observerIntersection.observe(element);
```
tsteur: I haven't looked too deep yet but if this callback is triggered multiple times, and the element is not in the viewport, does this mean we start observing the element every time meaning we call `observerIntersection.observe(element)` several times for the same element under circumstances?

Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -310,10 +294,86 @@
...
+
+                                    return;
+                                }
+
+                                var percentVisible = Math.max(getPercentVisible(element), minPercentVisible);
+                                commonTrigger(triggerEvent, percentVisible, element);
+                                if (fireTriggerWhen === 'oncePage') {
```
tsteur: fyi we have the similar code here in multiple places. Also around `if (self.scrollIndex) { TagManager.window.offScroll(self.scrollIndex);` etc. Maybe could be reused by adding a method?

tsteur (changes requested):
@AltamashShaikh looking already quite good 👍  left a few comments

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: Yes makes sense, updated as suggested

lang/en.json
AltamashShaikh: Added a note in description

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: Yes its not needed, removed

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: Yes, I have moved to a common function and should cover all cases, please check

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: Yes nice catch!
I added a fix to trigger only once

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: Updated code to execute callback every 120ms as discussed, please check

Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -272,47 +257,148 @@
...
+                if (observerIntersection) {
+                    observerIntersection.unobserve(node);
+                }
+            }
+        }
+
+        function getDynamicNodes(selectors) {
+            var dynamicNodes = [];
+            selectors.split(',').forEach(function (node) {
+                if (!documentAlias.querySelector(node)) {
```
tsteur: fyi we need to use `TagManager.dom.bySelector()` here to make sure things work across cross browsers. While all browsers might support this method that also support mutations it's still good to use a central logic for it just in case we need to change things there in the future etc.

Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -272,47 +257,148 @@
...
+                triggeredNodes.push(node); // to avoid possible memory leaks as much as possible we add onceElement only when needed
+                if (observerIntersection) {
+                    observerIntersection.unobserve(node);
+                }
+            }
+        }
+
+        function getDynamicNodes(selectors) {
+            var dynamicNodes = [];
+            selectors.split(',').forEach(function (node) {
```
tsteur: the variable here is `selector` not `node`?

Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -272,47 +257,148 @@
...
+                if (!documentAlias.querySelector(node)) {
+                    dynamicNodes.push(node.trim());
+                }
+            });
+
+            return dynamicNodes;
```
tsteur: do I see this right this actually returns a list of selectors not nodes?

Not 100% sure why we are doing this logic as later in `mutationObserverCallback` we query all selectors again anyway?

Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -272,47 +257,148 @@
...
+                    });
+
+                    observerMutation.observe(targetNode, config);
+                }
+            };
+        }
+
+        function mutationObserverCallback(mutationsList, triggerEvent) {
+            notInDomNodes.forEach(function (node) {
+                var element = documentAlias.querySelector(node);
```
tsteur: wouldn't we want to match all nodes in some cases? 
Even if `oncePerPage` or `oncePerElement` was selected then we'd still need to look over all nodes as one of them might be visible while others aren't? And it might not be the first one that is visible.

Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -272,47 +257,148 @@
...
                         }
-                    } else if (onlyOncePerElement) {
-                        observerIntersection.unobserve(entry.target);
-                        triggeredNodes.push(entry.target); // to avoid possible memory leaks as much as possible we add onceElement only when needed
+
+                        mutationObseverTimeout = true;
+                        setTimeout(function () {
+                            mutationObseverTimeout = false;
+                            mutationObserverCallback(mutationsList, triggerEvent);
+                            notInDomNodes = getDynamicNodes(selectors);
```
tsteur: not sure we can do this actually since a selector might match at a later point?

tsteur (changes requested):

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: Removed above function

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: Removed above function

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: Changed logic

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: Yes, removed

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: changed

Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -310,9 +273,122 @@
...
+                        }, 120);
+                    });
+
+                    observerMutation.observe(targetNode, config);
+                }
+            };
+        }
+
+        function mutationObserverCallback(mutationsList, triggerEvent) {
+            for (var index in mutationsList) {
```
tsteur, line 301: @AltamashShaikh can we execute `TagManager.dom.bySelector(selectors)` maybe only once at the beginning of the method here (before the `for` loop)?

Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -310,9 +273,122 @@
...
+
+                        mutationObseverTimeout = true;
+                        setTimeout(function () {
+                            mutationObserverCallback(allMutationsList, triggerEvent);
+                            allMutationsList = [];
+                            mutationObseverTimeout = false;
+                        }, 120);
+                    });
+
+                    observerMutation.observe(targetNode, config);
```
tsteur: not too important, here we could directly use `documentAlias.body`

Template/Trigger/ElementVisibilityTrigger.web.js
tsteur: this should make things a lot faster as these selector queries are very slow

tsteur (changes requested):
I haven't tested yet but code looks overall good @AltamashShaikh 🎉 . Hoping to give it a test tmrw. Left 2 minor comments for now

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: Updated as suggested

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: Yes changed as suggested, please check

Template/Trigger/ElementVisibilityTrigger.web.js
```diff
@@ -310,9 +273,125 @@
...
+                    });
+
+                    observerMutation.observe(documentAlias.body, config);
+                }
+            };
+        }
+
+        function mutationObserverCallback(mutationsList, triggerEvent) {
+            var domElements = [];
+            TagManager.dom.bySelector(selectors).forEach(function (element) {
```
tsteur: @AltamashShaikh no change needed but out of interest it would also work to use ` var domElements = TagManager.dom.bySelector(selectors)`?

tsteur (approved):
nice @AltamashShaikh left one more optional comment just to understand.

It worked nicely. I noticed one issue but I think this may have been an issue before. Created https://github.com/matomo-org/tag-manager/issues/362

Good to merge from my perspective.

Template/Trigger/ElementVisibilityTrigger.web.js
AltamashShaikh: Yes it would work, updated the code

Issue #217, opened 2020-01-24
Enable Debugging mode without URL parameter


peterbo:
The Debug mode doesn't seem to work without adding the mtmPreviewMode-Parameter to the URL. Enabling the Debug-Mode without the parameter would be nice, because often it is not possible to attach additional parameters (e.g. checkout funnel, security enhances sites with whitelisted parameter names, thank you pages, etc.)

tsteur:
BTW @peterbo when enabling the preview mode, it currently sets a cookie so you don't need the URL parameters. The cookie should be valid for one week. However, it doesn't work when Matomo runs on a different domain which is often the case. 

We'd basically need to find a way to set a cookie on a different frame which maybe could be achieved if a landing page has the Matomo tracking code embedded. It would need to be quite save though ensuring this feature cannot be misused.

I suppose we'd need to have some URL parameter `websiteUrl?enablePreviewMode=$containerUrl` where the format of the container name would need to be checked strictly. This would then maybe set the cookie? If there are multiple urls configured we would need to try and set the cookie on all urls?

peterbo:
Thanks for answering, Thomas! What mechanism did GTM base this on? Do they deploy a third party cookie? It seems like they deliver some kind of JSON debug configuration from the container-URL (which is the tracking server URL) if a debug & a session cookie is present and valid. But it seems to be pretty complex from what I can see. Probably an enhancement for later! Thank you!

Laul0:
I opened an issue related to this one in the wrong repo: matomo-org/matomo/issues/15941

Laul0:
In my case, if I enforce the domain into the cookie, the preview mode appears:

<img width="1367" alt="Screen Shot 2020-05-14 at 16 49 19" src="https://user-images.githubusercontent.com/15279205/81984879-76c65880-9603-11ea-9371-fb8fc4a8f5fc.png">

By default, the cookie domain is : `yourmatomo.domain.ext`

Edit this one with: `yourwebiste.domain.ext` and the preview appears.

> Note: the subdomains are different but the domain is the same in my case.

utrautmann:
It's not only an issue with different domains. In my case is the preview cookie set to the same domain like my website, but no preview panel appears.

But the preview cookie contains a path "/piwik" , because Matomo is installed in mydomain/piwik/

![grafik](https://user-images.githubusercontent.com/25771066/97107447-e1f27b80-16c7-11eb-852d-45f78abe6b65.png)

[...]

PR #355, opened 2021-09-01
Changes to enable debug mode without query param

AltamashShaikh:
### Description:

Changes to enable debug mode without query param.
Issue: #217 

### Review


tsteur:
@AltamashShaikh this plugin is one of very few (or the only plugin) where we actually don't need to maintain a changelog as the changes will be mentioned in the [Matomo changelog](https://matomo.org/changelog/) release. This will be generated by the release manager.

Controller.php
```diff
@@ -157,8 +157,11 @@ private function renderManageContainerTemplate($template, $variables = array())
 
                 $mtmPreviewId = PreviewCookie::COOKIE_NAME . '=' .$idContainer;
                 $mtmPreviewId = SafeDecodeLabel::decodeLabelSafe($mtmPreviewId);
+                $previewCookie = new PreviewCookie();
+                $debugSiteUrl = $previewCookie->getDebugSiteUrl();
                 $disableLink = '<a class="title disablePreviewDebug" onclick=\'tagManagerHelper.disablePreviewMode(' . json_encode($release['idcontainer']) . ')\'>' . Piwik::translate('TagManager_DisablePreview') .'</a>';
-                $notification = new Notification(Piwik::translate('TagManager_PreviewDebugEnabledNotification', array($version, '<strong>', '</strong>', '<strong>?' . $mtmPreviewId . '</strong>', '<strong>&' . $mtmPreviewId . '</strong>', $disableLink)));
+                $urlInput = '<br><div class="preview-debug-url-div"><input type="url" id="previewDebugUrl" class="browser-default" placeholder="https://yourdebugsite.com" value="'.$debugSiteUrl.'"> <button type="button" class="btn btn-small" onclick=\'tagManagerHelper.changeDebugUrl(' . json_encode($release['idcontainer']) . ',"'.$debugSiteUrl.'")\'>Debug</button></div>';
```
tsteur: could we make the text field a bit wider @AltamashShaikh ? the URL is otherwise mostly hidden
image

tsteur: Also can we move this part into a Twig file? You can then render this either using `renderTemplate()` or using `new View()...`;

The benefit is that you can escape things properly like the debug site url. In twig we can use different escaping strategies like `debugSiteUrl|e('html_attr')` and `debugSiteUrl|e('js')` etc to prevent XSS

tsteur: @mattab I'm assuming we would maybe want to prefill the URL with the configured main url (if any configured) ?
image

lang/en.json
```diff
@@ -163,6 +163,9 @@
         "DiffDeleted": "Deleted",
         "DisablePreview": "Disable Preview",
         "DisablingPreviewPleaseWait": "Disabling preview mode, please wait...",
+        "UpdatingDebugSiteUrlPleaseWait": "Updating debug site URL, please wait...",
+        "DebugUrlNoUrlErrorMessage": "Please Enter a URL to start debugging",
+        "DebugUrlSameUrlErrorMessage": "Already debugging the same site, please visit the %1$ssite%2$s or enter new url to debug",
```
tsteur: fyi here is a dot missing at the end. Same above in "DebugUrlNoUrlErrorMessage" and possibly others.

javascripts/tagmanager.js
```diff
@@ -1635,6 +1635,15 @@
                 throwError: throwError,
                 Container: Container,
                 addContainer: function (containerConfig, templates) {
+                    var setDebugFlag = urlHelper.getQueryParameter('setDebugFlag');
+                    if (setDebugFlag) {
+                        if (setDebugFlag == 1) {
+                            document.cookie = 'mtmPreviewMode=mtmPreview' + containerConfig.idsite + '_' + containerConfig.id + '%3D1;SameSite=Lax';
```
tsteur: I wonder if we should limit the time here ? I think the standard preview cookie restricts it to one week.

javascripts/tagmanager.js
```diff
@@ -1635,6 +1635,15 @@
                 throwError: throwError,
                 Container: Container,
                 addContainer: function (containerConfig, templates) {
+                    var setDebugFlag = urlHelper.getQueryParameter('setDebugFlag');
```
tsteur: btw we would probably want to prefix that URL parameter with `mtm` like `mtmSetDebugFlag`

tests/UI/expected-screenshots/ContainerVersion_version_some_exist_view_user.png.orig
```diff
@@ -0,0 +1,8 @@
+version https://git-lfs.github.com/spec/v1
```
tsteur: fyi here seems to be a merge conflict file which we can remove

javascripts/tagmanagerHelper.js
```diff
@@ -173,9 +174,36 @@
...
+        var id = 'TagManager_changeDebugSiteUrl';
+        var context = 'warning'; // or 'warning' or 'error' or 'success'
+        var UI = require('piwik/UI');
+        var notification = new UI.Notification();
+        if (!newUrl) {
+            return notification.show(_pk_translate('TagManager_DebugUrlNoUrlErrorMessage'), {context: context, id: id, title: ''});
+        } else if (newUrl === oldUrl) {
+            return notification.show(_pk_translate('TagManager_DebugUrlSameUrlErrorMessage', ['<a href="'+newUrl+'" target="_blank">','</a>']), {context: context, id: id, title: ''});
+        }
+        var piwikApi = piwikHelper.getAngularDependency('piwikApi');
```
tsteur: fyi usually when we're outside angular then we use the `ajaxHelper` class. It be usually fine to use `piwikApi` but as we're also in the process of removing angular it may be better to use `ajaxhelper` here so we won't need to migrate it later.

Like 

```

            var ajaxRequest = new ajaxHelper();
            ajaxRequest.addParams(ajaxParams, 'get');
            ajaxRequest.withTokenInUrl();
            ajaxRequest.setCallback(callback);
            ajaxRequest.setFormat('html');
            ajaxRequest.send();
```

tsteur (commented):
@AltamashShaikh I wonder now if we should always support only one domain.

What I did was first enter `https://www.example.com` as URL. Because that page doesn't embed one of our tag manager containers then when disabling the preview it never closed the window. There's probably no way around this though.

Then I entered another URL like `https://local.matomo.org` and whenever I enabled or disabled the preview mode then it would always still open example.com and never close it and I think there was no way for me to get rid of this wrong URL.

It then always opened 
image
but it also didn't seem to open my other URL.

javascripts/tagmanager.js
tsteur: BTW we also need to exclude this parameter from being tracked. Otherwise this URL parameter would appear in Matomo reports like `https://mydomain.com/path?mtmSetDebugFlag=1`. We can define such parameters to be excluded.

Controller.php
mattab: What would be the most useful from a user point of view? (what would 80% of users prefer to see in this field) @AltamashShaikh @tsteur

tsteur: I would say if only one domain is configured for a site, then we would prefill it with the main url, otherwise not prefill it. We would also prefill if they have configured the same domain twice for http and https like `http://www.example.com` and `https://www.example.com`. I probably wouldn't prefill if they configured `https://www.example.com` and `https://example.com`(this would be different subdomain). Most users will only have one domain and for these we could prefill? 
Other users can then still enter the URL (and path) if needed

AltamashShaikh:
> @AltamashShaikh I wonder now if we should always support only one domain.
> 
> What I did was first enter `https://www.example.com` as URL. Because that page doesn't embed one of our tag manager containers then when disabling the preview it never closed the window. There's probably no way around this though.
> 
> Then I entered another URL like `https://local.matomo.org` and whenever I enabled or disabled the preview mode then it would always still open example.com and never close it and I think there was no way for me to get rid of this wrong URL.
> 
> It then always opened
> image
> but it also didn't seem to open my other URL.

Can you check if popups are allowed ?
I was unable to replicate the same on my local

tsteur:
Popups are allowed I think. At least it opened the one for example.com but it didn't open the one for the other URL by the looks. And the closing probably just didn't work because the container wasn't embedded there.

Controller.php
AltamashShaikh: Made text field wider
Moved to twig

The URL is prefilled based on the current url we entered, for first time we get the site url using this below function
Site::getMainUrlFor($idSite);

lang/en.json
AltamashShaikh: Updated

javascripts/tagmanager.js
AltamashShaikh: Yes missed this, added a 7 day expiry

javascripts/tagmanager.js
AltamashShaikh: updated and excluded

tests/UI/expected-screenshots/ContainerVersion_version_some_exist_view_user.png.orig
AltamashShaikh: deleted

javascripts/tagmanagerHelper.js
AltamashShaikh: started using ajaxHelper

AltamashShaikh:
> Popups are allowed I think. At least it opened the one for example.com but it didn't open the one for the other URL by the looks. And the closing probably just didn't work because the container wasn't embedded there.

Okay can you try rebuilding the container, Else we can discuss this tomorrow

tests/UI/expected-screenshots/ContainerTag_edit_updated_back_to_list.png.orig
```diff
@@ -0,0 +1,8 @@
+version https://git-lfs.github.com/spec/v1
```
tsteur: @AltamashShaikh this file will need to be deleted there is a merge conflict

API.php
```diff
@@ -1174,6 +1175,7 @@ public function enablePreviewMode($idSite, $idContainer, $idContainerVersion = n
 
         $cookie = new PreviewCookie();
         $cookie->enable($idSite, $idContainer);
+        $cookie->enableDebugSiteUrl(Site::getMainUrlFor($idSite));
```
tsteur: @AltamashShaikh could we do this similar to mentioned in https://github.com/matomo-org/tag-manager/pull/355#discussion_r706949800 where we do this only if there is one domain configured?

tests/UI/expected-screenshots/ContainerTag_edit_updated_back_to_list.png.orig
AltamashShaikh: deleted

AltamashShaikh:
> @AltamashShaikh I wonder now if we should always support only one domain.
> 
> What I did was first enter `https://www.example.com` as URL. Because that page doesn't embed one of our tag manager containers then when disabling the preview it never closed the window. There's probably no way around this though.
> 
> Then I entered another URL like `https://local.matomo.org` and whenever I enabled or disabled the preview mode then it would always still open example.com and never close it and I think there was no way for me to get rid of this wrong URL.
> 
> It then always opened
> image
> but it also didn't seem to open my other URL.

There was a bug in disable, it was reading the siteMainUrl instead of input field
I fixed this can you please check again?

tsteur:
@mattab can you give this one a try and see if that works for you?

API.php
AltamashShaikh: As discussed we are removing this and will let the user to enter the debug url

tsteur:
@mattab can you have a look at this one? Ideally in next days so we can include it in 4.5 release otherwise 4.6

AltamashShaikh:
@tsteur Added changes suggested by @mattab 
Need to update FAQs and doc will update that too today.
Meanwhile can you have a look if the new changes looks okay to you

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## Changelog
 
 0.3.6
+
+* Added support to debug a site without appending the debug parameters in the URL
```
tsteur: Actually we can remove this changelog file @AltamashShaikh these changes will be mentioned instead in Matomo On-Premise changelog

javascripts/tagmanager.js
```diff
@@ -1635,6 +1635,17 @@
                 throwError: throwError,
                 Container: Container,
                 addContainer: function (containerConfig, templates) {
+                    var mtmSetDebugFlag = urlHelper.getQueryParameter('mtmSetDebugFlag');
+                    if (mtmSetDebugFlag) {
+                        if (mtmSetDebugFlag == 1) {
+                            var date = new Date();
+                            date.setTime(date.getTime() + (7 * 24 * 60 * 60 * 1000));
+                            document.cookie = 'mtmPreviewMode=mtmPreview' + containerConfig.idsite + '_' + containerConfig.id + '%3D1;expires=' + date.toUTCString() + ';SameSite=Lax';
```
tsteur: I wonder if it's maybe better in both cases the `encodeURIComponent()` the `idsite` and `container config ID`. I know they only have specific characters in there just wondering in case there's somehow an issue that the containerId could be changed might be good to have

javascripts/tagmanagerHelper.js
```diff
@@ -173,9 +173,47 @@
...
+        );
+        ajaxRequest.setFormat('html');
+        piwikHelper.modalConfirm('<h2>' + _pk_translate('TagManager_UpdatingDebugSiteUrlPleaseWait') + '</h2>', {});
+        ajaxRequest.send();
+    };
+    tagManagerHelper.updateDebugSiteFlag = function (url, idContainer, debugFlag) {
+        if (!url || !idContainer || !debugFlag) {
+            return;
+        }
+        window.open(url + (url.indexOf('?') == -1 ? '?' : '&') + 'mtmPreviewMode=' + idContainer + '&mtmSetDebugFlag=' + debugFlag, '_blank', 'noopener');
```
tsteur: can we also set `noreferrer` here next to `noopener` (technically noreferrer will also set noopener but might not hurt to mention both)?

tsteur: also here we will also url encode all the `idContainer` and the `debugFlag` just to be safe.

templates/previewDebugUrl.twig
```diff
@@ -0,0 +1,7 @@
+<br>
+<div class="preview-debug-url-div">
+    <input type="url" id="previewDebugUrl" class="browser-default" placeholder="https://yourdebugsite.com" value="{{ debugSiteUrl|e('html_attr') }}">
+    <button type="button" class="btn btn-small" onclick='tagManagerHelper.changeDebugUrl({{ idcontainer }},"{{ debugSiteUrl|e('html_attr') }}")'>
```
tsteur: fyi noticed when entering eg `foo` and then press debug then it will open `https://yourdomain.com/foo`. That can be actually good I suppose. Similar when entering `?er454534` into the text field etc.

templates/previewDebugUrl.twig
```diff
@@ -0,0 +1,7 @@
+<br>
+<div class="preview-debug-url-div">
+    <input type="url" id="previewDebugUrl" class="browser-default" placeholder="https://yourdebugsite.com" value="{{ debugSiteUrl|e('html_attr') }}">
```
tsteur: Let's maybe use a different domain just in case someone is registering that one and does something not good with it :)

In https://github.com/matomo-org/matomo/blob/4.x-dev/plugins/Installation/tests/UI/expected-screenshots/Installation_setup_website.png we use `https://example.org`

Not sure if it makes sense to set the placeholder as the main url if there is any for the current site. Maybe people would get confused and think it's already pre-filled?

tsteur (commented):
Nice, works @AltamashShaikh left a few comments.

BTW random thought. When a URL is entered, wonder if we would want to make the URL parameters `?mtmPreviewMode...` and `&mtm...` clickable as a link 
image

eg when entered both would in above example link to https://matomo.org?mtmPreviewMode=s73d1tB3 maybe?

AltamashShaikh:
> Nice, works @AltamashShaikh left a few comments.
> 
> BTW random thought. When a URL is entered, wonder if we would want to make the URL parameters `?mtmPreviewMode...` and `&mtm...` clickable as a link
> image
> 
> eg when entered both would in above example link to https://matomo.org?mtmPreviewMode=s73d1tB3 maybe?

Yes that would be much better, will update the code

CHANGELOG.md
AltamashShaikh: removed

javascripts/tagmanagerHelper.js
AltamashShaikh: set to noreferrer and added encodeURIComponent

templates/previewDebugUrl.twig
AltamashShaikh: Changed it to https://example.org and I think we should keep that only as it could be confusing if we keep the entered URL as the placeholder

Controller.php
```diff
@@ -157,8 +157,24 @@ private function renderManageContainerTemplate($template, $variables = array())
 
                 $mtmPreviewId = PreviewCookie::COOKIE_NAME . '=' .$idContainer;
                 $mtmPreviewId = SafeDecodeLabel::decodeLabelSafe($mtmPreviewId);
+                $previewCookie = new PreviewCookie();
+                $debugSiteUrl = $previewCookie->getDebugSiteUrl();
+                $mtmPreviewIdUrlStart = $mtmPreviewIdUrlEnd = '';
+                if (!empty($debugSiteUrl)) {
+                    $mtmPreviewIdUrlStart = '<a href="' . $debugSiteUrl . (stripos($debugSiteUrl, '?') !== False ? '&' : '?') . $mtmPreviewId . '" target="_blank">';
```
tsteur: @AltamashShaikh could we use Twig for this here to prevent any possible XSS?

javascripts/tagmanager.js
```diff
@@ -1635,6 +1635,19 @@
...
                 Container: Container,
                 addContainer: function (containerConfig, templates) {
+                    var mtmSetDebugFlag = urlHelper.getQueryParameter('mtmSetDebugFlag');
+                    if (mtmSetDebugFlag) {
+                        var idSite = encodeURIComponent(containerConfig.idsite);
+                        var containerID = encodeURIComponent(containerConfig.id);
+                        if (mtmSetDebugFlag == 1) {
+                            var date = new Date();
+                            date.setTime(date.getTime() + (7 * 24 * 60 * 60 * 1000));
+                            document.cookie = 'mtmPreviewMode=mtmPreview' + idSite + '_' + containerID + '%3D1;expires=' + date.toUTCString() + ';SameSite=Lax';
```
tsteur, line 1645: @AltamashShaikh can we urlencode these idsite and container parameters here? also in the `else` below? (not needed for date)

AltamashShaikh: Do you mean this?
https://github.com/matomo-org/tag-manager/blob/920fb90be784ba7ec289f2b66ade832d54501752/javascripts/tagmanager.js#L1640-L1648

tsteur: 👍

tsteur:
@AltamashShaikh I've just done a last test, when I enter `https://example.org?cx=3` as debug URL and then click on debug button, it opens `https://mymatomodomain.org/https%3A%2F%2Fexample.org%3Fcx%3D3?mtmPreviewMode=4epo2wMc&mtmSetDebugFlag=-1` instead of opening example.org...

To reproduce:

* Enter debug url `https://example.org`
* Click debug -> this seems to work
* Change the URL to `https://example.org?cx=3`
* Click debug -> opens wrong URL
* Enter debug url `https://example.org`
* Click debug -> seems to also no longer work

AltamashShaikh:
> @AltamashShaikh I've just done a last test, when I enter `https://example.org?cx=3` as debug URL and then click on debug button, it opens `https://mymatomodomain.org/https%3A%2F%2Fexample.org%3Fcx%3D3?mtmPreviewMode=4epo2wMc&mtmSetDebugFlag=-1` instead of opening example.org...
> 
> To reproduce:
> 
> * Enter debug url `https://example.org`
> * Click debug -> this seems to work
> * Change the URL to `https://example.org?cx=3`
> * Click debug -> opens wrong URL
> * Enter debug url `https://example.org`
> * Click debug -> seems to also no longer work

@tsteur Thanks for finding this bug :+1: 
It was due to wrong escaping, I updated the code

tsteur:
Looks good to merge @AltamashShaikh the only thing left I would do is to publish the FAQ you link to and update the link in this PR to link to the FAQ directly, not to the overview of links. This way if the link ever changes user will be redirected to correct FAQ.

Issue #225, opened 2020-02-18
Explain what the Tag Manager "Enable DNT" setting in the Matomo Configuration variable does


Findus23:
Inspired by a confusion in https://forum.matomo.org/t/browser-do-not-track-kills-active-opt-out-ability/36116

People get confused by the fact that they can enable DNT support both in Matomo and in Tag Manager. The tag manager setting should explain that it is just an addition client-side check that stops sending the data at all.

cherkavi:
With/Without "Support Do Not Track preference" Tag manager JS part (client) doesn't send any data to server.

Tag Manager Debug/Preview is showing in the Logs that "firing this tag" but no requests in browser log.

cherkavi:
Official docker container Matomo.
matomo              latest              4f05222e2406

Dry run working properly from browser: http://127.0.0.1:8080/matomo.php?idsite=4&rec=1
( it is nice to have the same for curl )

Error writing working properly:
```
docker exec -it apache_app_1 /bin/bash
sed --in-place  "s/define('PIWIK_PRINT_ERROR_BACKTRACE', true);/define('PIWIK_PRINT_ERROR_BACKTRACE', false);/g" /var/www/html/index.php
```

Debug flag is working as expected:
```bash
sed --in-place 's/debug = 1/debug = 0/g' /var/www/html/config/config.ini.php
```

UI ( JS ) code for sending "visits" and write down to table working properly
```sql
select * from matomo_log_visit
```

Tag manager ( Client part ) does not sending any requests to Matomo Server.
During the Debug/Preview sessions I see "fired events" but after "new version" and "publishing" and "switching off debug" - nothing

Issue #49, opened 2018-08-04
Also translate tag parameter descriptions


Findus23:
https://github.com/matomo-org/tag-manager/blob/cb9826810977a550f547514ffdef414b6aabdba3/Template/Tag/CustomHtmlTag.php#L39

At the moment the descriptions of tag parameters are not translated. Ideally there should be a default translation key similar to the tag description.

Findus23:
Interestingly there is 
https://github.com/matomo-org/tag-manager/blob/f5db47ffc05225ad25f4eeb92c22f503a731e107/lang/en.json#L133

which has another description (which is translated and shown in the UI), so maybe the only thing left to do is remove the unused hardcoded descriptions.

AltamashShaikh:
Fixed in this PR : https://github.com/matomo-org/tag-manager/pull/358

PR #358, opened 2021-09-03
Added description for matomo dnt setting, #225

AltamashShaikh:
### Description:

Added description for matomo dnt setting
Issue: #225 

### Review


Template/Variable/MatomoConfigurationVariable.php
```diff
@@ -93,6 +93,7 @@ public function getParameters()
             }),
             $this->makeSetting('enableDoNotTrack', false, FieldConfig::TYPE_BOOL, function (FieldConfig $field) {
                 $field->title = 'Enable Do Not Track';
+                $field->description = 'Enable this feature to not track users who opt out of tracking using Mozilla\'s (proposed) Do Not Track setting';
```
sgiehl: Looks good. @tsteur is there a special reason why all those fields & descriptions are not translatable. Or is that something that might be worth to change in this or another pull request?

Findus23:
Sorry, but I wasn't clear enough in my issue as that's not what I meant.
People know what DNT is, but they are confused why they can both configure it in Tag Manager and in their Matomo Config independently.

What I wanted to see explained is that the Matomo setting makes sure that the server drops all requests it receives with the DNT header while the tag manager setting does the same as adding `_paq.push(["setDoNotTrack", true]);` to your tracking code which means that Matomo will check DNT client-side and if it is enabled send no data to Matomo.

Template/Variable/MatomoConfigurationVariable.php
tsteur: there's #49

Template/Variable/MatomoConfigurationVariable.php
```diff
@@ -79,71 +81,72 @@ public function getParameters()
...
-                $field->title = 'Track All Content Impressions';
-                $field->description = 'Enables the content tracking feature by scanning the entire DOM for all content blocks and tracks all impressions once the page has loaded.';
+                $field->title = Piwik::translate('TagManager_MatomoConfigurationMatomoTrackAllContentImpressionsTitle');
+                $field->description = Piwik::translate('TagManager_MatomoConfigurationMatomoTrackAllContentImpressionsDescription');
             }),
             $this->makeSetting('trackVisibleContentImpressions', false, FieldConfig::TYPE_BOOL, function (FieldConfig $field) {
-                $field->title = 'Track Visible Content Impressions';
-                $field->description = 'Enables the content tracking feature by scanning the entire DOM for all content blocks but only tracks content impressions once the user scrolls to the content and the content is actually visible.';
+                $field->title = Piwik::translate('TagManager_MatomoConfigurationMatomoTrackVisibleContentImpressionsTitle');
```
Findus23, line 114: It's been a while since I last edited Tag Manager objects, but I think you don't need to specify this explicitly as the fallback already uses this key.

lang/en.json
```diff
@@ -553,6 +553,58 @@
         "WindowUnloadTriggerName": "Window Unload",
         "ZendeskChatTagDescription": "Reach your customers, instantly via web, mobile, and messaging.",
         "ZendeskChatTagHelp": "This tag allows you to add the Zendesk Chat form to your website.",
-        "ZendeskChatTagName": "Zendesk Chat (formerly Zopim)"
+        "ZendeskChatTagName": "Zendesk Chat (formerly Zopim)",
```
Findus23: It would be great if you could keep the en.json ordered as it makes diffs in the future nicer.

If you don't want to do it manually as that's annoying, you can use something like https://github.com/Findus23/matomo-utils/blob/main/sortTranslationKeys.py (comment out the lines that ignore plugins not by me)

Findus23: Thinking about it, we could even create a new console command that sorts all translation keys (ideally in the same method python uses) 
@sgiehl

Findus23 (changes requested):

lang/en.json
tsteur: @Findus23 feel free to create an issue for that and we can schedule it but be great to not work on it for now and having it sorted might not be that important. For example it can also be beneficial to have translation keys grouped by logic etc.

lang/en.json
```diff
@@ -553,6 +553,58 @@
...
+        "MatomoConfigurationMatomoIDSiteException": "The idSite can only include idSites and variables.",
+        "MatomoConfigurationMatomoEnableLinkTrackingTitle": "Enable Link Tracking",
+        "MatomoConfigurationMatomoEnableLinkTrackingDescription": "Enables the automatic download and outlink tracking.",
+        "MatomoConfigurationMatomoEnableCrossDomainLinkingTitle": "Enable Cross Domain Linking",
+        "MatomoConfigurationMatomoEnableCrossDomainLinkingDescription": "Enable this to accurately measure the same visitor across multiple domain names.",
+        "MatomoConfigurationMatomoEnableDoNotTrackTitle": "Enable Do Not Track",
+        "MatomoConfigurationMatomoEnableDoNotTrackDescription": "Enable this feature to not track users who opt out of tracking using Mozilla's (proposed) Do Not Track setting",
+        "MatomoConfigurationMatomoEnableDoNotTrackInlineHelp": "%1$sNote: This feature when enabled sets _paq.push([\"setDoNotTrack\", true]); which will not send any tracking request unlike \"Matomo Setting\" which drops the request if no tracking header is set and DNT is enabled%2$s",
```
tsteur: > %1$sNote: This feature when enabled sets _paq.push([\"setDoNotTrack\", true]); which will not send any tracking request unlike \"Matomo Setting\" which drops the request if no tracking header is set and DNT is enabled%2$

can we maybe make it like this:

> %1$sNote: When enabled it sets _paq.push([\"setDoNotTrack\", true]); which will not send any tracking request when a visitor has DNT enabled unlike the \"DNT Matomo Setting\" which drops the request on the server when a visitor has DNT is enabled%2$

tsteur (commented):
Left a minor suggestion, can merge otherwise 👍

Template/Variable/MatomoConfigurationVariable.php
AltamashShaikh: Moved to en.json to make it translatable

lang/en.json
AltamashShaikh: Yes updated.

lang/en.json
AltamashShaikh: Updated the en.json file.

PR #389, opened 2021-12-10
Started deleting variable if entity recursion exception is thrown on import

### LINKED ISSUE #388: Fixed error when importing a container: This variable cannot be deleted as it is used in other places
image

* Create a new container. 
* Go to versions
* Click on import
* Import below data

```
{"idcontainer":"lRjDHEmG","idsite":1,"context":"web","name":"test2","description":"","created_date":"2021-12-09 19:28:11","updated_date":"2021-12-09 19:28:11","created_date_pretty":"Dec 9, 2021 20:28:11","updated_date_pretty":"Dec 9, 2021 20:28:11","revision":0,"version":{"name":"","description":"","revision":0,"created_date":"2021-12-09 19:28:11","created_date_pretty":"Dec 9, 2021 20:28:11","updated_date":"2021-12-09 19:28:11","updated_date_pretty":"Dec 9, 2021 20:28:11"},"tags":[],"triggers":[],"variables":[{"idvariable":36,"type":"MatomoConfiguration","name":"Matomo Configuration","parameters":{"matomoUrl":"{{Matomo Configuration}}","idSite":"1","enableLinkTracking":true,"enableCrossDomainLinking":false,"enableDoNotTrack":false,"enableJSErrorTracking":false,"enableHeartBeatTimer":false,"trackAllContentImpressions":false,"trackVisibleContentImpressions":false,"disableCookies":false,"requireConsent":false,"requireCookieConsent":false,"setSecureCookie":false,"cookieDomain":"","cookiePath":"","cookieSameSite":"Lax","domains":[],"alwaysUseSendBeacon":false,"userId":"","customDimensions":[],"bundleTracker":true,"registerAsDefaultTracker":true,"jsEndpoint":"matomo.js","trackingEndpoint":"matomo.php"},"lookup_table":[],"default_value":"","created_date":"2021-12-09 19:28:53","updated_date":"2021-12-09 19:28:53","created_date_pretty":"Dec 9, 2021 20:28:53","updated_date_pretty":"Dec 9, 2021 20:28:53"}]}
```

What this import represents (refs https://github.com/matomo-org/tag-manager/pull/293 ):

> Eg create a MatomoConfiguration variable named "Matomo Configuration" and set Matomo URL to "https://matomo.org{{Matomo Configuration}}". 

The above one there is a variable that references itself which is causing this issue. Technically, this should not be happening anymore as we prevented a while ago this from happening and any newer container it shouldn't be possible to configure such a container. Unless someone configured the JSON themselves.

Maybe we could time box and look into this for 1 hour to see if we can fix this issue and make such imports happen and if not we might put it back in the backlog. If we can't solve it quick, maybe in import we could catch this error and then attach an extra message so it reads like `This variable cannot be deleted as it is used in other places. To remove this variable, first remove all places where this variable is used. It could be that a variable references itself within a variable.`. We wouldn't change the error message though in other places.

AltamashShaikh:
Started deleting variable if entity  recursion exception is thrown on import.
Fixes: #388 

### Description:

Please include a description of this change and which issue it fixes. If no issue exists yet please include context and what problem it solves.

### Review

### DIFF
--- a/API/Import.php
+++ b/API/Import.php
@@ -146,8 +146,7 @@ class Import
                     'lookupTable' => $variable['lookup_table'],
                 ));
             } catch (EntityRecursionException $e){
-                //Delete the variable from container if recursion exception is thrown
-                $this->variables->deleteContainerVariable($idSite, $idContainerVersion, $variable['idvariable']);
+                throw new \Exception(Piwik::translate('TagManager_EntityRecursionExceptionForVariable', array($variable['name'] . '(' . $variable['type'] . ')')));
             }
         }
 
--- a/API/Import.php
+++ b/API/Import.php
@@ -9,6 +9,7 @@ namespace Piwik\Plugins\TagManager\API;
 
 use Piwik\API\Request;
 use Piwik\Piwik;
+use Piwik\Plugins\TagManager\Exception\EntityRecursionException;
 use Piwik\Plugins\TagManager\Input\AccessValidator;
 use Piwik\Plugins\TagManager\Model\Container;
 use Piwik\Plugins\TagManager\Model\Tag;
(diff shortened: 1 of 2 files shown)


tsteur:
@AltamashShaikh I see, the problem is when adding the variable there is a recursion. I thought it might happen when deleting or so. I think simply deleting the variable may not be expected as the user wouldn't be made aware of this problem that we skipped a variable and it could maybe even cause issues if other triggers or tags use this variable.

Instead we'd want to make sure that the user gets an accurate error message. In that case I would have expected we see the message like `'It seems an entity references itself or a recursion is caused in some other way. It may be related due to these entites: "'.implode(',', $entries). '". Please check if the entity references itself maybe or if a recursion might happen in another way.`. Do we know why we aren't seeing this?

This be the most important for now.

As a next step we would need to see if we can revert to the previously active draft as the import would then have worked partially I assume? This is out of scope for now unless it's super trivial. Let's mostly make sure that the correct error message is shown. We could then keep the issue open and eventually work on the other part.

API/Import.php
```diff
@@ -133,16 +134,20 @@ public function importContainerVersion($exportedContainerVersion, $idSite, $idCo
...
+                    'idContainer' => $idContainer,
+                    'idContainerVersion' => $idContainerVersion,
+                    'type' => $variable['type'],
+                    'name' => $variable['name'],
+                    'parameters' => $variable['parameters'],
+                    'defaultValue' => $variable['default_value'],
+                    'lookupTable' => $variable['lookup_table'],
+                ));
+            } catch (EntityRecursionException $e){
+                throw new \Exception(Piwik::translate('TagManager_EntityRecursionExceptionForVariable', array($variable['name'] . '(' . $variable['type'] . ')')));
```
AltamashShaikh, line 149: @tsteur Should we capture all such exceptions in a Array lets say and based on that throw an exception or 1 at a time is fine ?

tsteur: not needed for now @AltamashShaikh

PR #412, opened 2022-01-24
Added option to filter only fired tags with indicator

### LINKED ISSUE #340: Add option to filter events in Preview mode
## Context:
Currently when creating triggers (For example an `All Elements Click`) that has a filter, an event is logged in the Preview mode for every trigger regardless of if the trigger condition was met or not or if it fired a tag or not.

This can cause issues, especially when users create more than a few triggers. It's not unreasonable for users that are tracking even a single website or application with the Tag Manager to have 20 or more triggers for various things. If for example all 20 of them were `All Elements Click` those 20 triggers would create 20 events in the Preview Mode for every single click and forces the user to click through every single event to find the one that fired a tag.

This is definitely the expected behaviour in a lot of cases, especially when troubleshooting or testing new triggers in order to determine if the Trigger is being detected correctly by the tag manager regardless of if the filter was matched or not (To determine if the cause is due to a filter or other incorrectly configured setting)

This issue is further compounded when users use a single container to track multiple websites, with each website having triggers and tags that were created specifically for each website.

## Possible solution:
Something like a toggle or checkbox that users can use to hide events that didn't fire any tags (And potentially the other way around - a toggle to show only events that didn't trigger any tags may be useful in some cases)

AltamashShaikh:
### Description:

Added option to filter only fired tags with indicator in preview debug mode.
Fixes: #340 

### Review

### DIFF
--- a/templates/debug.twig
+++ b/templates/debug.twig
@@ -16,16 +16,18 @@
             margin-top: 0;
         }
 
-        .tm-icon-ok {
-            position: absolute;
-            background-repeat: no-repeat;
+        .tm-icon {
             display: inline-block;
-            background-size: cover;
-            margin-left: 0.5rem;
-            margin-top: 2px;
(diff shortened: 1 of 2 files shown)


AltamashShaikh:
Screenshot from 2022-01-25 12-38-14
Screenshot from 2022-01-25 12-37-46

@tsteur Does this looks good to you ?

templates/debug.twig
```diff
@@ -74,13 +93,13 @@
...
+                        <span class="item" style="font-weight: normal;"> Events <br><br> <input type="checkbox" class="onlyFiredTags-chk" name="onlyfiredTags" value="1" ng-model="onlyfiredTags"> Only fired tags</span>
                     </li>
                     <li ng-show="(mtmDebug.mtmEvents|length) == 0">No event executed</li>
                     <li class="menuTab" role="menuitem" ng-class="{'active': eventIndex === mtmDebug.selectedEventIndex}"
-                    ng-repeat="event in mtmDebug.mtmEvents | orderBy:'$index':true">
+                    ng-repeat="event in mtmDebug.mtmEvents | orderBy:'$index':true" ng-hide="!event.tags.length && onlyfiredTags">
                         <a class="item ng-binding"  ng-click="mtmDebug.selectEvent(event.index0)" title="Time: {{ '{{ event.time }}'|raw }}. Trigger: '{{ '{{ event.metTrigger.name }}'|raw }}'">
-                            {{ '{{ event.index }}'|raw }}: {{ '{{ event.name }}'|raw }}
+                            {{ '{{ event.index }}'|raw }}: {{ '{{ event.name }}'|raw }} <span class="{{ '{{ event.tags.length ? \'icon-ok\' : \'icon-error\' }}' }}"></span>
```
AltamashShaikh: Screenshot from 2022-01-25 12-42-13

@tsteur 
Using the existing icons from Matomo core there is 1 problem, if the Matomo instance is on https and the tracking website is on http the icon doesn't seems to load due to CORS.

Should we add a base64 encoded svg as an inline CSS to solve this ?

tsteur: sounds good 👍

tsteur (changes requested):
@AltamashShaikh this worked for me. 

While no event is executed yet there is the "No event executed" text misaligned 
image

Also if any possible be great to trigger the checkbox when clicking on the label "Only fired tags". So far I have to click on the checkbox itself to toggle it.
image

On hover we also want to explain what the tick icon means. And for simplicity let's maybe not show an icon if it wasn't triggered. I mean instead of 
image

it would be this: 
image

AltamashShaikh:
@tsteur Made changes as suggested, please check

templates/debug.twig
```diff
@@ -69,18 +94,34 @@
...
+                        </span>
                     </li>
-                    <li ng-show="(mtmDebug.mtmEvents|length) == 0">No event executed</li>
+                    <li ng-show="(mtmDebug.mtmEvents|length) == 0" style="padding: 0 0 1rem 1.2rem;">No event executed</li>
                     <li class="menuTab" role="menuitem" ng-class="{'active': eventIndex === mtmDebug.selectedEventIndex}"
-                    ng-repeat="event in mtmDebug.mtmEvents | orderBy:'$index':true">
+                    ng-repeat="event in mtmDebug.mtmEvents | orderBy:'$index':true" ng-hide="!event.tags.length && onlyfiredTags">
                         <a class="item ng-binding"  ng-click="mtmDebug.selectEvent(event.index0)" title="Time: {{ '{{ event.time }}'|raw }}. Trigger: '{{ '{{ event.metTrigger.name }}'|raw }}'">
-                            {{ '{{ event.index }}'|raw }}: {{ '{{ event.name }}'|raw }}
+                            {{ '{{ event.index }}'|raw }}: {{ '{{ event.name }}'|raw }} <span title="Events with fired tags" ng-show="event.tags.length">
```
tsteur: can we rename the title to `This tag was fired` ?

tsteur (changes requested):
Works nicely @AltamashShaikh 

Left one more comment.

And can we adjust the https://github.com/matomo-org/tag-manager/blob/4.x-dev/LEGALNOTICE#L45-L55 to mention that we are using an icon in `templates/debug.twig`? Not sure from which font it is?

templates/debug.twig
AltamashShaikh: Updated

AltamashShaikh:
> Works nicely @AltamashShaikh
> 
> Left one more comment.
> 
> And can we adjust the https://github.com/matomo-org/tag-manager/blob/4.x-dev/LEGALNOTICE#L45-L55 to mention that we are using an icon in `templates/debug.twig`? Not sure from which font it is?

Its under Icomoon - Free
Screenshot from 2022-01-28 08-56-48

Please check if I have mentioned it correctly or not

tsteur:
That looks good @AltamashShaikh 👍

AltamashShaikh:
@tsteur looks good to merged right ?

tsteur:
Yes 👍

PR #431, opened 2022-02-16
Added changes for tawktotag to work with new widgetID changes

### LINKED ISSUE #426: Added changes for tawktotag to work with new widgetID changes
The embed script for the Tawk To service is no longer working:

[](https://github.com/matomo-org/tag-manager/blob/4.x-dev/Template/Tag/TawkToTag.web.js)

The embed URL now contains a second Widget ID parameter. For example:

`https://tawk.to/chat/<Property ID>/<Widget Id>`

If the "default" parameter is used, a 404 response is returned.

`s1.src = 'https://embed.tawk.to/' + tawkToId + '/default';`

AltamashShaikh:
### Description:

Added changes for tawktotag to work with new widgetID changes
Fixes: #426

### Review

### DIFF
--- a/Template/Tag/TawkToTag.php
+++ b/Template/Tag/TawkToTag.php
@@ -35,6 +35,12 @@ class TawkToTag extends BaseTag
                 $field->validators[] = new NotEmpty();
                 $field->validators[] = new CharacterLength(16, 30); // we limit to 30 so users don't accidentally enter a 32 digit API key
             }),
+            $this->makeSetting('tawkToWidgetId', '', FieldConfig::TYPE_STRING, function (FieldConfig $field) {
+                $field->title = 'tawk.to Widget ID';
+                $field->description = 'You can get the Widget ID by logging into Tawk.to, going to "Administration" and clicking on "Chat Widget" and selecting an appropriate widget. The widget ID can be retrieved from "Direct Chat Link" as https://tawk.to/chat/{SITE_ID}/{WIDGET_ID}.';
+                $field->validators[] = new NotEmpty();
+                $field->validators[] = new CharacterLength(9, 20); // we limit to 20 so users don't accidentally enter a 32 digit API key
+            }),
         );
     }
 
--- a/Template/Tag/TawkToTag.php
+++ b/Template/Tag/TawkToTag.php
@@ -35,11 +35,11 @@ class TawkToTag extends BaseTag
                 $field->validators[] = new NotEmpty();
                 $field->validators[] = new CharacterLength(16, 30); // we limit to 30 so users don't accidentally enter a 32 digit API key
             }),
-            $this->makeSetting('tawkToWidgetId', '', FieldConfig::TYPE_STRING, function (FieldConfig $field) {
+            $this->makeSetting('tawkToWidgetId', 'default', FieldConfig::TYPE_STRING, function (FieldConfig $field) {
                 $field->title = 'tawk.to Widget ID';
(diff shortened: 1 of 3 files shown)


Template/Tag/TawkToTag.php
```diff
@@ -35,6 +35,12 @@ public function getParameters() {
                 $field->validators[] = new NotEmpty();
                 $field->validators[] = new CharacterLength(16, 30); // we limit to 30 so users don't accidentally enter a 32 digit API key
             }),
+            $this->makeSetting('tawkToWidgetId', '', FieldConfig::TYPE_STRING, function (FieldConfig $field) {
+                $field->title = 'tawk.to Widget ID';
+                $field->description = 'You can get the Widget ID by logging into Tawk.to, going to "Administration" and clicking on "Chat Widget" and selecting an appropriate widget. The widget ID can be retrieved from "Direct Chat Link" as https://tawk.to/chat/{SITE_ID}/{WIDGET_ID}.';
+                $field->validators[] = new NotEmpty();
+                $field->validators[] = new CharacterLength(9, 20); // we limit to 20 so users don't accidentally enter a 32 digit API key
```
AltamashShaikh: @tsteur 
Couldn't find much on the length of widget ID
I tried setting the widget name to a very long text and the id was of length:9 only for safety I kept it till 20

AltamashShaikh: Also please check the description if the user can understand it or not

Template/Tag/TawkToTag.php
```diff
@@ -35,6 +35,12 @@ public function getParameters() {
                 $field->validators[] = new NotEmpty();
                 $field->validators[] = new CharacterLength(16, 30); // we limit to 30 so users don't accidentally enter a 32 digit API key
             }),
+            $this->makeSetting('tawkToWidgetId', '', FieldConfig::TYPE_STRING, function (FieldConfig $field) {
```
tsteur: fyi I've just checked in an old account we had, there it was still `default`. So to not break it for existing users, I assume setting `default` as default value might help. Otherwise we risk that it worn't work anymore for customers that have their account already for a while?

I then created another new widget in our account, and there it did use two different parts indeed. The second part was 9 characters long. Allowing up to 20 sounds good.

Template/Tag/TawkToTag.web.js
```diff
@@ -1,10 +1,12 @@
 (function () {
     return function (parameters, TagManager) {
+        var Tawk_API = Tawk_API || {}, Tawk_LoadStart = new Date();
         this.fire = function () {
             var tawkToId = parameters.get("tawkToId");
+            var tawkToWidgetId = parameters.get("tawkToWidgetId");
             var s1 = document.createElement("script"), s0 = document.getElementsByTagName("script")[0];
             s1.async = true;
-            s1.src = 'https://embed.tawk.to/' + tawkToId + '/default';
+            s1.src = 'https://embed.tawk.to/' + tawkToId + '/' + tawkToWidgetId;
```
tsteur: can we encodeURIComponent both parameters here?

tsteur (changes requested):

Template/Tag/TawkToTag.web.js
AltamashShaikh: Yes added

Template/Tag/TawkToTag.php
AltamashShaikh: I have updated it to default and reduced the minLength limit to 7 otherwise it will fail as default is of 7 characters.

Template/Tag/TawkToTag.php
AltamashShaikh: @tsteur I was wondering should we make the description same for both the fields ? As we can get both the details from 1 link

PR #432, opened 2022-02-16
[Vue] migrate debugging angularjs controller to vue component

diosmosis:
### Description:

Changes:
* migrate debugging angularjs controller to vue component

### Review


diosmosis:
Note: UI test failure looks random.

Also, note that this will break existing previews since the HTML appears to be saved on the filesystem. Disabling and enabling the preview fixes it. If needed we can perhaps disable all previews in an update in core.

AltamashShaikh:
> Note: UI test failure looks random.
> 
> Also, note that this will break existing previews since the HTML appears to be saved on the filesystem. Disabling and enabling the preview fixes it. If needed we can perhaps disable all previews in an update in core.

@diosmosis 
We should update it in core, since users might complain that after a recent upgrade Matomo Tag Manager preview is not working.
Also, we can regenerate all the containers or make use of this method `generateContainerIfHasPreviewRelease` to regenerate only when preview enabled.

AltamashShaikh (changes requested):
@diosmosis
We should update it in core, since users might complain that after a recent upgrade Matomo Tag Manager preview is not working.
Also, we can regenerate all the containers or make use of this method generateContainerIfHasPreviewRelease to regenerate only when preview enabled.

diosmosis:
@AltamashShaikh created a core PR here: https://github.com/matomo-org/matomo/pull/18921

AltamashShaikh (approved):
@diosmosis I just saw this
https://github.com/matomo-org/tag-manager/blob/4.x-dev/TagManager.php#L55

On update we are triggering `regenerateReleasedContainers` so we don't need the change in core, we can keep the plugin changes but no need to make changes in core

diosmosis:
@AltamashShaikh oh I see, that's interesting, I'll close the core PR, nice catch, thanks!

PR #449, opened 2022-03-11
Added disableBrowserFeatureDetection in MatomoConfigurationVariable

### LINKED ISSUE #443: Add new option to Matomo Variable to disable browser feature detection in Matomo Tag
See https://matomo.org/faq/how-do-i-disable-browser-feature-detection-completely/

We need a new checkbox that when enabled, would call the method `disableBrowserFeatureDetection` in the Matomo Tag.

This will help people stay privacy compliant with TTDSG if they want/need to.

AltamashShaikh:
### Description:

Added disableBrowserFeatureDetection in MatomoConfigurationVariable
Fixes: #443

### Review


tsteur (changes requested):
@AltamashShaikh see issue description, we also need to use the variable to call the right method on the tracker in the MatomoTag.

Because the method to disable browser feature detection was only added at a later point, we should maybe also check if the method exists on the tracker before calling it.

AltamashShaikh:
> @AltamashShaikh see issue description, we also need to use the variable to call the right method on the tracker in the MatomoTag.
> 
> Because the method to disable browser feature detection was only added at a later point, we should maybe also check if the method exists on the tracker before calling it.

Yes, updated as suggested

tests/System/expected/test_webContext__TagManager.getAvailableVariableTypesInContext.xml
```diff
@@ -982,6 +982,22 @@
...
+						<name>disableBrowserFeatureDetection</name>
+						<title>Disable Browser Feature Detection</title>
+						<value>0</value>
+						<defaultValue>0</defaultValue>
+						<type>boolean</type>
+						<uiControl>checkbox</uiControl>
+						<uiControlAttributes>
+						</uiControlAttributes>
+						<availableValues />
+						<description>Disable detection of visitor's current browser resolution and browser plugins supported.</description>
```
tsteur: We could say here: `By default, Matomo accesses some information from the visitor’s browser like the browser resolution. Some privacy regulations may only allow accessing information that is stored on a visitor’s device after having consent. If this applies to you, you can disable this feature.`.

The inline help we can remove.

Don't know if we can put links in here? I believe it maybe be possible in the inline help to use HTML.  If any possible be great to link to the FAQ: https://matomo.org/faq/how-to/how-do-i-disable-browser-feature-detection-completely/

Like `Learn more`...

tsteur (changes requested):
Works 👍 
@AltamashShaikh in the variable can we move this setting to just after the "Same Site Cookie" setting and before the "Domains" setting? This way we have all the relevant cookie settings together.

In the Matomo Tag itself, can we call the disable browser features method before `disableCookies`? Just to prevent that any of the other methods that we call use the browser features.  Checked in https://github.com/matomo-org/matomo/blob/4.8.0/js/piwik.js#L7126 and this is what we currently do in the tracker itself where we define in which order the methods should be called when using `_paq.push` (which we aren't using here)

tests/System/expected/test_webContext__TagManager.getAvailableVariableTypesInContext.xml
AltamashShaikh: @tsteur I modified the text to `By default, Matomo accesses some information from the visitor’s browser like the browser resolution. Some privacy regulations may allow access of such information from the visitor’s device only after having a consent. If this applies to you, you can disable this feature.`

Yes we can put links using `inlineHelp`, added a link to FAQ.

Also moved the variable and and calling of method in Variable and Tag as suugested

Template/Tag/MatomoTag.web.js
```diff
@@ -117,38 +117,43 @@
...
+                        tracker.disableCookies();
                     }
 
                     if (matomoConfig.enableCrossDomainLinking) {
                         tracker.enableCrossDomainLinking();
                     }
 
-                    if (matomoConfig.setSecureCookie) {
-                        tracker.setSecureCookie(true);
+                    if (matomoConfig.requireConsent) {
```
tsteur: @AltamashShaikh assuming we are now calling all the methods in the same order as in the tracker file, we would maybe need to call `requireConsent` last? In https://github.com/matomo-org/matomo/blob/4.8.0/js/piwik.js#L7126 it shows up after the cookie domain

Template/Variable/MatomoConfigurationVariable.php
```diff
@@ -166,6 +166,11 @@ public function getParameters()
                     'Strict' => 'Strict',
                 );
             }),
+            $this->makeSetting('disableBrowserFeatureDetection', false, FieldConfig::TYPE_BOOL, function (FieldConfig $field) {
+                $field->title = Piwik::translate('TagManager_MatomoConfigurationMatomoDisableBrowserFeatureDetectionTitle');
+                $field->description = Piwik::translate('TagManager_MatomoConfigurationMatomoDisableBrowserFeatureDetectionDescription');
+                $field->inlineHelp = Piwik::translate('TagManager_MatomoConfigurationMatomoDisableBrowserFeatureDetectionInLineHelp', array('<br><strong>', '<a href="https://matomo.org/faq/how-to/how-do-i-disable-browser-feature-detection-completely/" target="_blank">', '</a>', '</strong>'));
```
tsteur: not sure why, the "target='_blank'" doesn't seem to be in the HTML for some reason. Not sure if anything is stripping this or what is happening. It opened the link in the same window. Maybe Vue is removing it? We wouldn't spend too much time on this one so we could otherwise just create a bug report in core that attributes are being ignored.

also what we good but I think no longer required is to add `rel="noreferrer noopener"` attribute.

image

tsteur (changes requested):

Template/Tag/MatomoTag.web.js
AltamashShaikh: Moved, now calling before `enableDoNotTrack` and after `enableLinkTracking`

Template/Variable/MatomoConfigurationVariable.php
AltamashShaikh: I had a word with dizzy on this, its due to recent vue changes, `it needs a rel="noreferrer noopener" otherwise its considered insecure` I have added the rel attribute and that seems to work

tsteur (approved):

MatomoForumNotifications:

https://forum.matomo.org/t/mtm-and-disablebrowserfeaturedetection/45405/2

PR #453, opened 2022-03-11
Allow 0 value for trigger comparison

### LINKED ISSUE #450: Fix bug preventing "0" value for trigger condition
Good morning,
I'd like to use "0" or "1" as conditions but I get an error no matter which type I choose (equals, equals exactly, regex):
Bildschirmfoto 2022-03-11 um 09 00 23

sgiehl:
### Description:

fixes #450

### Review

### DIFF
--- a/Validators/TriggerConditions.php
+++ b/Validators/TriggerConditions.php
@@ -67,7 +67,7 @@ class TriggerConditions extends BaseValidator
                 throw new Exception(Piwik::translate('TagManager_ErrorArrayMissingValue', array('comparison', $titlePlural, $index)));
             }
 
-            if (!is_string($condition['expected']) || 0 === strlen($condition['expected'])) {
+            if (!isset($condition['expected']) || !is_string($condition['expected']) || 0 === strlen($condition['expected'])) {
                 throw new Exception(Piwik::translate('TagManager_ErrorArrayMissingValue', array('expected', $titlePlural, $index)));
             }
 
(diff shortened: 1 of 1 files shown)


Validators/TriggerConditions.php
```diff
@@ -67,7 +67,7 @@ public function validate($value)
                 throw new Exception(Piwik::translate('TagManager_ErrorArrayMissingValue', array('comparison', $titlePlural, $index)));
             }
 
-            if (empty($condition['expected'])) {
+            if (!is_string($condition['expected']) || 0 === strlen($condition['expected'])) {
```
AltamashShaikh: @sgiehl this would still throw an exception for such case
https://3v4l.org/qkEcG

AltamashShaikh:
```suggestion
            if (empty($condition['expected']) && strlen($condition['expected'])===0) {
```

sgiehl: @AltamashShaikh But shouldn't the parameter actually always be a string, as it's coming from API?
Your suggestions actually wouldn't work either for `null` value, as PHP 8 would throw an error when passing `null` to `strlen`.

AltamashShaikh: yes it will fail for null values, I can see the values are always coming as string so it should be good to merge this change

AltamashShaikh (approved):
Good to merge, just check the test cases once

AltamashShaikh (changes requested):
we need to add !isset check too, due to which the test case seems to be failing

Validators/TriggerConditions.php
AltamashShaikh:
```suggestion
            if (!isset($condition['expected'])) || !is_string($condition['expected']) || 0 === strlen($condition['expected'])) {
```

AltamashShaikh (approved):

PR #467, opened 2022-03-31
CustomHTML fix html ordering issue when position is bodyStart

### LINKED ISSUE #466: The output is difference from input (Custom HTML tag)
I created a tag using the Custom HTML tag, I entered this code as you read it:

```
<script src="https://cdn.jsdelivr.net/gh/dixonandmoe/rellax@master/rellax.min.js"></script>
<script>
  var rellax = new Rellax('.rellax');
</script>
```

The output is different, the two scripts have change order and that means the script wont work:

```
<script> var rellax = new Rellax('.rellax'); </script>
<script src="https://cdn.jsdelivr.net/gh/dixonandmoe/rellax@master/rellax.min.js"></script>
```

AltamashShaikh:
### Description:

CustomHTML fix html ordering issue when position is bodyStart.
Fixes: #466

### Review


Template/Tag/CustomHtmlTag.web.js
```diff
@@ -101,25 +101,28 @@
...
-                newScript.async = element.async;
+            if (element.hasAttribute('async')) {
+                newScript.setAttribute('async',element.getAttribute('async'));
+            }
+            if (element.hasAttribute('onload')) {
+                newScript.setAttribute('onload', element.getAttribute('onload'));
             }
 
-            newScript.type = "text/javascript";
+            newScript.setAttribute('type', 'text/javascript');
```
tsteur: fyi here we might actually also want to check `getAttribute('type')`?

Template/Tag/CustomHtmlTag.web.js
```diff
@@ -141,11 +144,14 @@
...
-        function insertNode(parent, child, append)
+        function insertNode(parent, child, append, previousElement)
         {
             if (append || !parent.firstChild) {
-                parent.appendChild(child);
+                return parent.appendChild(child);
             } else {
+                if (previousElement) {
+                    return previousElement.parentNode.insertBefore(child, previousElement.nextSibling);
+                }
```
tsteur, line 155: would we return an element as well when `parent.insertBefore` is executed?

Template/Tag/CustomHtmlTag.web.js
```diff
@@ -155,23 +161,24 @@
             var limit = 5000; // prevent endless loop
             var counter = 0;
             var child;
+            var previousElement = '';
```
tsteur: not too important but maybe we want to set null or something? considering we're dealing with HTML elements below and initialising a string, this may be misleading what information previousElement contains

Template/Tag/CustomHtmlTag.web.js
```diff
@@ -155,23 +161,24 @@
...
 
             while (counter in children && children[counter] && counter < limit) {
                 child = children[counter];
                 counter++;
 
                 if (isJavaScriptElement(child)) {
                     // we have to re-create the element, otherwise wouldn't be executed
-                    insertNode(parent, cloneScript(child), append);
+                    previousElement = insertNode(parent, cloneScript(child), append);
```
tsteur: none of the attributes are yet passing the previousElement to `insertNode`?

tsteur (commented):
fyi I know it's not ready for review yet but had a quick look as you pinged me in Slack. Overall that strategy looks like it'll work 👍

Template/Tag/CustomHtmlTag.web.js
AltamashShaikh: Yes we should do that, added same

Template/Tag/CustomHtmlTag.web.js
AltamashShaikh: I missed this, updated

Template/Tag/CustomHtmlTag.web.js
AltamashShaikh: Null is better

Template/Tag/CustomHtmlTag.web.js
AltamashShaikh: I actually pushed to show how the changes would look and was experimenting without `previousElement` and that got pushed by mistakenly, added back

AltamashShaikh:
@tsteur Looks good for review

tsteur:
@AltamashShaikh I just tried it with the example from the user in the issue. Getting this JS error 

image

with this configuration: 

image

same for BodyStart. Is this expected?

The dom is in the right order 👍 
image

executing `  var rellax = new Rellax('.rellax');` manually works.

AltamashShaikh:
> @AltamashShaikh I just tried it with the example from the user in the issue. Getting this JS error
> 
> image
> 
> with this configuration:
> 
> image
> 
> same for BodyStart. Is this expected?
> 
> The dom is in the right order +1 image
> 
> executing ` var rellax = new Rellax('.rellax');` manually works.

@tsteur Yes even I get the same error and I think its due to some limitation with appending js
`$('body').prepend('<script src="https://cdn.jsdelivr.net/gh/dixonandmoe/rellax@master/rellax.min.js"></script> <script> var rellax = new Rellax(\'.rellax\'); </script>');`
This throws the same error.

The ideal solution for this is to use onload something like this
```
<script src="https://cdn.jsdelivr.net/gh/dixonandmoe/rellax@master/rellax.min.js" onload="rellaxAfterLoad()"></script>
<script>
var rellax;
function rellaxAfterLoad() {
    rellax = new Rellax('.rellax');
}

</script>
```

tsteur:
@AltamashShaikh works 👍  and LGTM from my perspective if tests pass.

However, I wouldn't close the original issue as the problem still persists that it doesn't quite behave as expected and user would need to use "onload". We can still merge though as it already improves things a bit. Do you know how GTM handles this re the JS error? Maybe when no onload handler is defined, then we would need to detect this use case and only add the next script node once the "onload" or "onerror" event has been triggered? Not sure if GTM takes care of this or not. Anyway, it probably be still good to merge this already now and see this as a separate problem. We just don't want to close the issue the user created since it's still not really working.

AltamashShaikh:
> @AltamashShaikh works +1 and LGTM from my perspective if tests pass.
> 
> However, I wouldn't close the original issue as the problem still persists that it doesn't quite behave as expected and user would need to use "onload". We can still merge though as it already improves things a bit. Do you know how GTM handles this re the JS error? Maybe when no onload handler is defined, then we would need to detect this use case and only add the next script node once the "onload" or "onerror" event has been triggered? Not sure if GTM takes care of this or not. Anyway, it probably be still good to merge this already now and see this as a separate problem. We just don't want to close the issue the user created since it's still not really working.

I tried debugging whats in GTM, I can see GTM always adds to bottom and the code is also similar to us
Create a div
Add the customHTML as innerHTML of div
Parse childnodes and create an array
If element is script convert it to scriptElement via document.createElement('script');
Append the element to body

but if added via GTM above code works as it is, tried applying the changes but still not working
May be we can merge this and keep the issue open and work on it later

tsteur:
sounds good 👍

tsteur:
could mention in the issue if someone else experiences this issue to comment, and explain a workaround until it's fixed

PR #470, opened 2022-04-04
Added code to move debug window from top to bottom

### LINKED ISSUE #271: Tag Manager: Change the location of Preview window
Possibility of changing the location of Preview window.

**Request from a user:** We are working on measuring a chatbot. I use Matomo Tag Manager and the preview/debug functionality to test the triggers. However, because there is no option to minimize/resize the preview/debug mode i am not able to use this option on the chatbot. Since the chatbot is located at the right-bottom of the page (and the preview/debug mode overlays this).

AltamashShaikh:
### Description:

Added code to move debug window from top to bottom.
Fixes: #271 

### Review


tsteur:
@AltamashShaikh just gave it a test. I actually tested it a few times and thought the preview mode is not working anymore as I couldn't see the preview window after opening the window. Tried to debug and then noticed I have to scroll now to the bottom of the page to see the preview window instead of it sitting at the bottom "fixed" at any time. 

I think we'd still want to show the preview window like before "fixed" at the bottom by default. Otherwise it's hard to see what happens when you click somewhere and there be a lot of scrolling up and down involved etc.

We would then maybe ideally move it only in the visible viewport eg move the preview window further up or down but never out of sight/viewport.

And ideally we would maybe remember the position where in the viewport we had the preview pane last time and position it there again.

Is there a chance to show the "click here to move" on top of the existing iframe content to not hide more content on the current site? It could be just an icon that we overlay on top.

not too important but for some reason I can't move it fully to the top 
image

It might be even good enough to just have a button "move to top" and when it's in top "move to button". This may be less of an effort instead of having actually to drag and drop and might be more productive with just a single click. (or have it as an additional option, could be also just an icon). It can be maybe frustrating otherwise having to move it around too often.

AltamashShaikh:
@tsteur The dynamic change of position with viewport was creating issues so I have removed it completely.
Currently added "Stick to Top/Stick to Bottom".
On click we are refreshing the page as we have to prepend for top and append for bottom

tsteur:
@AltamashShaikh that works nicely. However, can we remove the page reload in any way? 

It can be frustrating if there's a page reload happening and in case there was a post request previously, information could get lost.

AltamashShaikh:
> @AltamashShaikh that works nicely. However, can we remove the page reload in any way?
> 
> It can be frustrating if there's a page reload happening and in case there was a post request previously, information could get lost.

@tsteur 
We could leverage `customEvent` and try to call renderPreviewFrame again but this will lead to loss of existing history of debug window.

In `previewmode.js`
```
document.addEventListener('ReloadDOMContent', function (e) {
  var iframe = document.getElementById('mtmDebugFrame');
  if (iframe) {
    iframe.remove();
  }
  renderPreviewFrame(e.detail.theContent, true);
});
```

In `Debugging.vue` on click add this
```
mtmUpdateDebugPosition() {
      const sevenDays = 7 * 60 * 60 * 24 * 1000;
      const currentCookieValue = getCookie(cookieName);
      const cookieValue = (currentCookieValue === 'top' ? 'bottom' : 'top');
      setCookie(cookieName, cookieValue, sevenDays);

      const content = document.getElementById('mtmDebugFrame').contentDocument ? document.getElementById('mtmDebugFrame').contentDocument.documentElement.innerHTML : '';
      var ReloadDOMContent = new CustomEvent("ReloadDOMContent", {
        "detail": {theContent:content}
      });

      document.dispatchEvent(ReloadDOMContent);
    }
```

Will the lost of history be fine ?

tsteur:
@AltamashShaikh Sorry I don't understand why we would need this?

Can we have a `div` like this 
image
which on click would then change the position of the iframe?

AltamashShaikh:
@tsteur `position:sticky` was not working as expected when trying to change `top/bottom` style. To fix this I have made the position as `fixed` instead of `sticky`.
This approach works but it act as a overlay on top/bottom elements due to which initial elements gets hidden below the debug iframe.

Eg: With position: fixed
Screenshot from 2022-04-19 08-38-53

Eg: With position: sticky
Screenshot from 2022-04-19 08-39-14

javascripts/previewmode.js
```diff
@@ -7,7 +7,29 @@ function htmlToElement(content)
...
+    var iframeClass = stickToTop() ? 'mtmStickyTop' : 'mtmStickyBottom';
+    return htmlToElement('<iframe class="'+ iframeClass +'" id="mtmDebugFrame" src="about:blank" frameborder="0" style="background-color:#edecec !important; clip: initial !important; display: inline !important; height:33% !important; opacity: 1 !important; visibility:visible !important; padding:0 !important; margin:0 !important; border:0 !important; border-top: 2px solid #fff !important; position:fixed !important; left:0 !important; width:100% !important; z-index:999999999 !important;min-height: 18rem;"></iframe>');
+}
+
+function stickToTop() {
+    return getCookie('mtmPreviewPosition') === 'top';
```
tsteur: we could potentially remove the entire `getCookie` method and shorten this to `document.cookie.indexOf('mtmPreviewPosition=top') > -1`?

javascripts/previewmode.js
```diff
@@ -28,18 +50,21 @@ function renderPreviewFrame(theContent)
     if (!previewFrame) {
         // might already exist when embedding multiple containers
         previewFrame = createDebugFrame();
-        document.body.appendChild(previewFrame);
+        var sheet = document.createElement('style');
```
tsteur, line 36: I wonder if this can cause issues when those styles are blocked by a content security policy like `Content-Security-Policy: style-src https://example.com/` see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src

Can we simply set the CSS on the iframe directly maybe like 
image

actually we can ignore it because we're already setting the style attribute meaning if the `style` attribute works, then this should work too. we can ignore it.

tsteur (commented):
Nice! That works really well now. The only thing I noticed is when it sticks to the top, then the `border-top` needs to be changed to `border-bottom` (`2px solid ...`). And left one other comment to simplify the cookie access

javascripts/previewmode.js
AltamashShaikh: Simple and does the work :+1:

AltamashShaikh:
@tsteur 
Simplified the cookie check and added `border-bottom` when `stickyTop` and `border-top` when `stickyBottom`

tsteur (approved):
Works nicely @AltamashShaikh 👍 💯

PR #477, opened 2022-05-04
Allow dynamic height on Tag manager preview

korve:
### Description:

This commit adds a feature that makes the preview bar of the tag manager vertically resizable. It adds a handle with which the user can adjust the height of the bar. The height is also saved in a cookie and restored when reloading the preview site.

I've implemented this because i had massive problems using the preview mode on my daily job. I never had enough real estate to view all tags, triggers etc. I hope this makes the preview bar more usable.

See the following video to get a idea of what this commit provides.

https://user-images.githubusercontent.com/4459994/166812796-fdefb896-59cf-44f1-bcb0-012a634880f1.mp4

### Review

### DIFF
--- a/javascripts/previewmode.js
+++ b/javascripts/previewmode.js
@@ -8,7 +8,7 @@ function htmlToElement(content)
 function createDebugFrame()
 {
     var iframeClass = stickToTop() ? 'mtmStickyTop' : 'mtmStickyBottom';
-    return htmlToElement('<iframe class="'+ iframeClass +'" id="mtmDebugFrame" src="about:blank" frameborder="0" style="cursor: row-resize !important; background-color:#edecec !important; clip: initial !important; display: inline !important; height: ' + getHeight() + '; opacity: 1 !important; visibility:visible !important; padding:0 !important; margin:0 !important; position:fixed !important; left:0 !important; width:100% !important; z-index:999999999 !important;min-height: 18rem;"></iframe>');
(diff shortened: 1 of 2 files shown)


AltamashShaikh:
@korve Thanks for the PR seems to work as expected.

AltamashShaikh:
@tsteur Can you have a look once?

tsteur:
@AltamashShaikh  I didn't review the code but from a UX perspective that works for me 👍  nice feature @korve and thanks a lot for the contribution 🎉 .

They only thing I was wondering if should maybe keep a higher min-height of the debug window to make sure some portion of it is always visible? Otherwise people might by accident change the min height, and then it's barely visible and they are thinking it's not shown at all anymore when in fact it is. That could reduce some possible frustrations and support emails.

AltamashShaikh:
@korve can we limit the resize to a certain height to make it always visible ?

korve:
> @korve can we limit the resize to a certain height to make it always visible ?

Yep, i'll look into it.

korve:
I think i've managed to limit the height. Currently it's set to 10% of the viewport.

I had to implement some feature testing to support IE < 11. Specifically `pointerdown` and `pointermove` events are not supported  and also `setPointerCapture` is not supported which allows capturing mousemove events outside the browser window.

AltamashShaikh (approved):
@korve Looks good :+1:

PR #489, opened 2022-06-02
Fixed circular dependency in debugging.vue

### LINKED ISSUE #487: TypeError: Converting circular structure to JSON in Angluar apps in Preview Mode
Same type of issue as #185 
However, this is happening on the most recent version of Matomo (4.10.0).
This issue seems to only be happening on sites that use Angular (PHP sites for example do not have the issue using the same container).

image

The user was able to reproduce the error by first clicking a button on their site and then clicking on the Event in the preview mode iFrame:
image

This results in a blank preview page and the **above** error in the console:
image

I've asked the user to confirm if the same was happening in 4.9 or if it only started after the update to 4.10.0
Will add a comment here if I get confirmation about that.

AltamashShaikh:
### Description:

Fixed circular dependency in debugging.vue
Fixes: #487

### Review


AltamashShaikh:
If someone needs to test the actual failing behaviour do let me know, as we got a URL to reproduce this bug but the person who shared the URL is not comfortable in sharing on public forum

vue/src/Debugging/Debugging.vue
```diff
@@ -170,7 +170,8 @@
               <tbody>
               <tr>
                 <td style="word-break: break-all">
-                  {{ selectedEvent?.eventData && JSON.stringify(selectedEvent.eventData) }}
+                  {{ selectedEvent?.eventData &&
+                  JSON.stringify(selectedEvent.eventData, getCircularReplacer) }}
```
diosmosis: @AltamashShaikh I haven't checked this, but before I do a further review, does it work if the code is just:

`{{ selectedEvent?.eventData }}`

?

AltamashShaikh: @diosmosis It works for non angular sites, but it doesn't work for angular sites

AltamashShaikh: I have shared the details on slack on how to reproduce it.

AltamashShaikh: @diosmosis I tried using `{{ selectedEvent?.eventData }}` the error goes away but it prints incorrectly
Screenshot from 2022-06-03 10-25-44

we need to stringify the JSON and also to avoid circular dependency we need the method, the issue is not because of angular site only, its evident in such sites due to `__ngContext__` having such circular dependency and can happen on any sites if they have such dependency due to triggered event

diosmosis: Ok. In that case I would say the code's fine except instead of calling `JSON.stringify(..., getCircularReplacer())` in the template on every update, it should be in computed properties, eg:

```
computed: {
  selectedEventData() {
    return this.selectedEvent?.eventData && JSON.stringify(this.selectedEvent.eventData, getCircularReplacer());
  },
},
```

then in the template:

```
{{ selectedEventData }}
```

The code looks fine otherwise, though I didn't test it locally.

AltamashShaikh: @diosmosis updated as suggested

diosmosis (approved):
FYI, getCircularReplacer() can be a free function now, it doesn't have to be a method. Otherwise the code looks fine. Pinging @sgiehl in case he wants to do his own review.

AltamashShaikh:
> FYI, getCircularReplacer() can be a free function now, it doesn't have to be a method. Otherwise the code looks fine. Pinging @sgiehl in case he wants to do his own review.

@diosmosis Updated to free function

PR #515, opened 2022-07-21
Added javascript masking

### LINKED ISSUE #503: Add support for `data-matomo-mask` attribute on events
Currently the data-matomo-mask works for the heatmap and session recording but it does not work when it comes to events.
The typical use case is the following one, if you set the tracking of your events on any clicks happening on your website, then, a user may click on a personal data (for example his login while logged), as a result, the {{click text}} information may be sent to Matomo.
It would be useful to set the data-matomo-mask in order to not have the {{click text}} information sent to Matomo.

snake14:
### Description:

Added javascript masking
Fixes: #503 

### Review


AltamashShaikh (changes requested):
@snake14 Congratulations on your first PR
We are capturing clickText for following triggers
1. AllDownloadsClickTrigger
2. AllElementsClickTrigger
3. AllLinksClickTrigger
4. DomElementVariable

It would be nice to make changes in [tagmanager.js](https://github.com/matomo-org/tag-manager/blob/4.x-dev/javascripts/tagmanager.js#L741) as it will cover all such case.

Also it will be great if you read about how we do masking in Heatmap plugins - from [here](https://developer.matomo.org/guides/heatmap-session-recording/setup)

I have shared few links on slack with you to check various masking scenarios, we have nested levels to like
```
<div class="parent" data-matomo-mask>
<button>Should be masked</button>
</div>
```
In the above case when we click on button, it will mask the clickText, as the parent has `data-matomo-mask` attribute

snake14:
@AltamashShaikh Thank you. I'll look into it some more. I wasn't sure if the Heatmap code applied here since what I saw appeared to insert CSS to do the masking. That's not an option with TagManager recorded events is it? Good point about nested levels. I'll see what I can do.

javascripts/tagmanager.js
```diff
@@ -754,6 +760,11 @@
                     return '';
                 },
                 getElementAttribute: function (node, attributeName) {
+                    // If the attribute belongs to a masked element, return a masked string.
+                    if (node && TagManager.dom.shouldElementBeMasked(node)) {
```
AltamashShaikh: Why have we added masking here ?

AltamashShaikh (changes requested):
@snake14 
We need to support `data-matomo-unmask` attribute to
```
<div id="d1" data-matomo-mask="true">
<span id="s1">Masked</span>
<span id="s2" data-matomo-unmask="true">
  <span id="s2_1">NO masking</span>
  <span id="s2_2" data-matomo-mask>masking</span>
</span>
</div>

<div id="d2">
<span id="s3" data-matomo-mask="true">Masked</span>
<span id="s4" data-matomo-unmask="true">No Masked</span>
</div>
```
Probably check for `data-matomo-unmask` should fix most of the cases
Check https://jsfiddle.net/9tjo438s/3/ for some examples

AltamashShaikh:
@snake14 I will also be good to add few test cases for `getElementText` with `data-matomo-mask` and `data-matomo-unmask` attributes, so that we can be sure in future that we haven't regressed anything.
Refer - https://github.com/matomo-org/tag-manager/blob/4.x-dev/tests/javascript/index.php#L632

javascripts/tagmanager.js
snake14: @AltamashShaikh I added that since there might be Variables that expose data-* attributes that contain sensitive information in Dimensions.

Like this:
image

With the code in the getElementAttribute function:
image

AltamashShaikh: @snake14 Not sure if we should mask attributes too.
@tsteur Do you think we should add masking of attributes ?

snake14:
@AltamashShaikh Thank you. I wasn't sure if I needed to implement the unmask attribute, but I'll look into adding support for that and I'll see what I can add to the tests.

javascripts/tagmanager.js
```diff
@@ -742,6 +742,17 @@
...
                         return;
                     }
+
+                    // If the content belongs to a masked element and it doesn't have any children, return a masked string.
+                    if (TagManager.dom.shouldElementBeMasked(node) && node.children.length === 0) {
+                        return '*******';
+                    }
+
+                    // If the element has children that should be masked, deal with that.
+                    if (TagManager.dom.elementHasMaskedChild(node)) {
```
AltamashShaikh, line 752: @snake14 Why do we need this ? Above code will do the needful correct ?

javascripts/tagmanager.js
```diff
@@ -754,6 +765,11 @@
                     return '';
                 },
                 getElementAttribute: function (node, attributeName) {
+                    // If the attribute belongs to a masked element, return a masked string.
+                    if (node && TagManager.dom.shouldElementBeMasked(node)) {
```
AltamashShaikh: I don't think we need this as `data-matomo-mask` only maskes the content of the node not the attributes content, It might add confusion for few people and I don't think its needed here

javascripts/tagmanager.js
```diff
@@ -953,6 +969,64 @@
                           callback(event, 'right');
                         }
                     }, true)
+                },
+                shouldElementBeMasked: function (element) {
```
AltamashShaikh, line 978: The logic here should cover all the cases

AltamashShaikh (changes requested):
@snake14 You can remove masking for `attributes` as we do not mask attributes in heatmap, same should be followed here.
`elementHasMaskedChild` is actually not needed, I see you have added for tests, can't we directly check the childrens by ID ? Something like this ?
```
strictEqual(true, dom.shouldElementBeMasked(document.getElementById('maskedDiv1')), 'Should mask maskedDiv1')
```

javascripts/tagmanager.js
snake14: @AltamashShaikh After looking more at the example you provided and the code in the heatmap plugin, I ended up handling masking part of the content of an element instead of everything. This is especially important if a div is masked, but one of its children has the unmask attribute. So, lines 747-749 handle masking elements that should be masked and don't have any children. The if statement on line 752 checks if the element has children elements that need to be masked.

javascripts/tagmanager.js
snake14: @AltamashShaikh If you're certain, I can remove it. Since the code that handles [Variable attributes](https://github.com/matomo-org/tag-manager/blob/bf6dd14fa70048c52236e833288fe3e66be32302/Template/Variable/DomElementVariable.web.js#L29) doesn't allow password inputs, I thought that maybe masking should mask attributes too.

AltamashShaikh: @snake14 my bad
I just checked we do masking for following attributes
`title`, `alt`, `label`, `placeholder`, `value`

we should add similar check here for above attributes

javascripts/tagmanager.js
snake14: @AltamashShaikh Could you please elaborate? This function accepts an element and returns whether or not the element needs to be masked. What else should be be doing?

javascripts/tagmanager.js
snake14: @AltamashShaikh Thank you. Only masking a specific group of attributes is a good idea. I'll make that change.

javascripts/tagmanager.js
AltamashShaikh: @snake14 Understood now, we need this change

snake14:
@AltamashShaikh I hadn't thought about simply passing the ID instead of the element. I could make that change if you like. My biggest hesitation is that would feel like it's following a different pattern than the existing functions, like addEventListener, getElementText, getElementClassNames, and getElementAttribute that accept the element as the argument and not the ID.

AltamashShaikh:
@snake14 I revisited the code after your comment, and now we just need to mask only few attributes not all and rest looks fine :+1:

javascripts/tagmanager.js
tsteur: Yes, we would mask some common attributes as it was implemented like `title` or `alt` 👍

AltamashShaikh:
@snake14 Looks good :tada: Nice work :+1: 
Also we should update/add the documentation for `getElementText, getElementAttribute, shouldElementBeMasked, elementHasMaskedChild and getElementTextWithMaskedChildren` [here](https://developer.matomo.org/api-reference/tagmanager/javascript-api-reference) 

This the repository [URL](https://github.com/matomo-org/developer-documentation) for documentation you need to setup on your local and checkout a new branch from `live` branch
For pre existing methods like `getElementText and getElementAttribute` we can mention that since Matomo 4.12, this method will mask the value if `data-matomo-mask` attribute is set

snake14:
@AltamashShaikh Thank you for all of your help. I will update the documentation.

snake14:
@AltamashShaikh I created the following [pull request](https://github.com/matomo-org/developer-documentation/pull/648) to document the changes.

AltamashShaikh (approved):

PR #519, opened 2022-07-27
Prevent possible XSS via changeDebugUrl

AltamashShaikh:
### Description:

Changes to fix possible XSS via changeDebugUrl

### Review

### DIFF
--- a/API.php
+++ b/API.php
@@ -1224,7 +1224,7 @@ class API extends \Piwik\Plugin\API
      * @param string $idContainer  The id of a container, for example "6OMh6taM"
      * @param string $url  The url to enable debug
      */
-    public function changeDebugUrl($idSite, $idContainer, $url)
+    public function changeDebugUrl($idSite, $url)
     {
         $this->accessValidator->checkWriteCapability($idSite);
         if (!filter_var($url, FILTER_VALIDATE_URL)) {
--- a/API.php
+++ b/API.php
@@ -37,6 +37,7 @@ use Piwik\Plugins\TagManager\Template\Trigger\TriggersProvider;
 use Piwik\Plugins\TagManager\Template\Variable\MatomoConfigurationVariable;
 use Piwik\Plugins\TagManager\Template\Variable\VariablesProvider;
 use Exception;
+use Piwik\UrlHelper;
 
 /**
  * API for plugin Tag Manager.
@@ -1227,7 +1228,12 @@ class API extends \Piwik\Plugin\API
     public function changeDebugUrl($idSite, $url)
     {
         $this->accessValidator->checkWriteCapability($idSite);
-        if (!filter_var($url, FILTER_VALIDATE_URL) || stripos($url, 'http') !== 0) {
+        if (
+            !filter_var($url, FILTER_VALIDATE_URL)
+            || stripos($url, 'http') !== 0
+            || !UrlHelper::isLookLikeSafeUrl($url)
+            || !UrlHelper::isLookLikeUrl($url)
+        ) {
             throw new Exception(Piwik::translate('TagManager_InvalidDebugUrl'));
         }
 
--- a/API.php
+++ b/API.php
@@ -1227,7 +1227,7 @@ class API extends \Piwik\Plugin\API
     public function changeDebugUrl($idSite, $url)
     {
         $this->accessValidator->checkWriteCapability($idSite);
-        if (!filter_var($url, FILTER_VALIDATE_URL)) {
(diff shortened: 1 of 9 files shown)


templates/previewDebugNotification.twig
```diff
@@ -5,7 +5,9 @@
     {% set mtmPreviewIdUrlStart = '' %}
     {% set mtmPreviewIdUrlEnd = '' %}
 {% endif %}
-
+<div piwik-alert="danger" id="debug-notification-alert" style="display: none">
+    <strong>Error!</strong> <span id="debug-notification-alert-message"></span>
```
sgiehl: Any reason for not having `Error!` translatable?

AltamashShaikh: I missed it, added translation

snake14 (approved):
The changes look good to me. Which issue is this for again? I don't see the issue number in the PR description or any of the commit messages. Did you do that for a specific reason?

AltamashShaikh:
> The changes look good to me. Which issue is this for again? I don't see the issue number in the PR description or any of the commit messages. Did you do that for a specific reason?

@snake14 This was a fix for XSS vulnerability report which I have shared in the slack channel you can check that

snake14:
> @snake14 This was a fix for XSS vulnerability report which I have shared in the slack channel you can check that

@AltamashShaikh Thank you. I see it now.

javascripts/tagmanagerHelper.js
```diff
@@ -227,22 +228,35 @@
             return tagManagerHelper.updateDebugSiteFlag(newUrl, idContainer, 1);;
         }
         var ajaxRequest = new ajaxHelper();
+        var template = $('<h2>' + _pk_translate('TagManager_UpdatingDebugSiteUrlPleaseWait') + '</h2>');
         ajaxRequest.addParams({
             module: 'API',
+            format: 'json',
             method: 'TagManager.changeDebugUrl',
             idContainer: idContainer,
```
sgiehl: guess this line can be removed as well, as the parameter isn't needed.
```suggestion
```

AltamashShaikh: @sgiehl removed

API.php
```diff
@@ -1224,8 +1224,13 @@ public function disablePreviewMode($idSite, $idContainer)
      * @param string $idContainer  The id of a container, for example "6OMh6taM"
      * @param string $url  The url to enable debug
      */
-    public function changeDebugUrl($idSite, $idContainer, $url)
+    public function changeDebugUrl($idSite, $url)
     {
+        $this->accessValidator->checkWriteCapability($idSite);
+        if (!filter_var($url, FILTER_VALIDATE_URL)) {
```
sgiehl: I'm actually wondering if using `FILTER_VALIDATE_URL` only might cause problems here. It does not check if the protocol is "safe". It could also be `ssh://` or similar. So we could additionally check `preg_match('/^https?:\/\//')`, or do you think other protocols are needed there as well?

AltamashShaikh: @sgiehl Added `stripos($url, 'http')` check to ensure valid protocol is present.

sgiehl: the problem with `stripos === false` is, that you don't check if it's in the beginning. So something like `ssh://http.webhost.com` would pass here... So guess you would need to compare with `!== 0`

AltamashShaikh: Good catch :+1:

AltamashShaikh: Updated and added few testcases too

API.php
```diff
@@ -1224,8 +1224,13 @@ public function disablePreviewMode($idSite, $idContainer)
      * @param string $idContainer  The id of a container, for example "6OMh6taM"
      * @param string $url  The url to enable debug
      */
-    public function changeDebugUrl($idSite, $idContainer, $url)
+    public function changeDebugUrl($idSite, $url)
     {
+        $this->accessValidator->checkWriteCapability($idSite);
+        if (!filter_var($url, FILTER_VALIDATE_URL) || stripos($url, 'http') !== 0) {
```
tsteur: fyi @AltamashShaikh does it make sense to also look at `UrlHelper::isLookLikeSafeUrl()` here additionally? it includes a few additional checks. There's also `isLookLikeUrl()` which may be interesting too not sure.

AltamashShaikh: @tsteur Ideally the 'stripos($url, 'http') === 0' should fix most of the cases but adding both the methods makes no harm.

tsteur: 👍 that's what I thought. Better be safe. Also in the future someone might add more changes to core and we won't remember to add them here as well meaning then we're covered.

PR #525, opened 2022-07-28
Updated so that javascript that's part of customHtml tags uses the no…

### LINKED ISSUE #326: Fix for Tagmanager working along with CSP when triggering a HTML tag
Customer's explanation of the issue:

I create a nonce for the actual embed Script, read: the regular matomo one, and the embed script for the tag manager. These two codes work perfectly fine. No csp issues arrise here.
Now lets say, i create a tag, that for the sake of simplicity, fires on every click to the page. The tag it self is a "CUSTOM HTML" tag, that simply says: '<script>console.log("Hello from the Tag")</script>' and gets appended to the end of the html body.

When i now click on the webpage the tag gets fired and i get a csp viloation. "[Report Only] Refused to execute inline script because it violates the following Content Security Policy directive ...". From my understanding this is correct, as a script gets inserted that neither has a nonce nor a hash and therefore can not be deemed to be trustworthy. The tag-manager tho, gets loaded via a nonce and therefore is trustworthy, i'd now need a mechanism to add the nonce from the tagmanager to script elements that get created by the matomo tag manager too. And this is what i struggle to do. Adding it as a custom-var does not work (read from a header eg) as the tagmanagment script removes most attributes of scripts added.

Sure, I could use script hashes, but then i'd need to know every script i'll add beforehand and add its hash to my CSP-Header, which defeats the purpose of the tag management.

snake14:
### Description:

Updated the Tag javascript file so that when a customHtml tag includes javascript, it checks if the main script has a nonce and copies it to the custom javascript so that it works with CSP.
Fixes: #326 
@tsteur This is a security issue, so I thought it would be good to have you look at it and make sure I'm not creating a vulnerability.

### Review

### DIFF
--- a/Template/Tag/CustomHtmlTag.web.js
+++ b/Template/Tag/CustomHtmlTag.web.js
@@ -127,9 +127,12 @@
 
             // If this script is executing using a nonce, set the nonce on its children too.
             var scriptWithNonce = TagManager.dom.bySelector('script[nonce]');
-            if (Array.isArray(scriptWithNonce) && scriptWithNonce.length) {
+            if (scriptWithNonce.length && scriptWithNonce[0]) {
                 scriptWithNonce = scriptWithNonce[0]; // Grab the first element.
-                newScript.nonce = scriptWithNonce.nonce;
+                // Try using the attribute first for compatibility, then go to the property.
+                var nonceAttr = TagManager.dom.getElementAttribute(scriptWithNonce, 'nonce');
+                nonceAttr = nonceAttr ? nonceAttr : scriptWithNonce.nonce;
+                newScript.setAttribute('nonce', nonceAttr);
             }
 
             return newScript;
--- a/Template/Tag/CustomHtmlTag.web.js
+++ b/Template/Tag/CustomHtmlTag.web.js
@@ -126,8 +126,9 @@
             }
 
             // If this script is executing using a nonce, set the nonce on its children too.
-            var scriptWithNonce = document.querySelector('script[nonce]');
-            if (typeof scriptWithNonce !== 'undefined') {
+            var scriptWithNonce = TagManager.dom.bySelector('script[nonce]');
+            if (Array.isArray(scriptWithNonce) && scriptWithNonce.length) {
+                scriptWithNonce = scriptWithNonce[0]; // Grab the first element.
                 newScript.nonce = scriptWithNonce.nonce;
             }
 
--- a/Template/Tag/CustomHtmlTag.web.js
+++ b/Template/Tag/CustomHtmlTag.web.js
@@ -125,6 +125,12 @@
                 newScript.setAttribute('type', element.getAttribute('type'));
             }
 
+            // If this script is executing using a nonce, set the nonce on its children too.
+            var scriptWithNonce = document.querySelector('script[nonce]');
+            if (typeof scriptWithNonce !== 'undefined') {
+                newScript.nonce = scriptWithNonce.nonce;
+            }
+
             return newScript;
         }
 
--- a/tests/javascript/index.php
+++ b/tests/javascript/index.php
@@ -2549,7 +2549,10 @@
             fireTemplateTag(templateToTest, params);
             var addedStyle4 = document.getElementById('customStyleTag4');
             strictEqual('var x = {};', addedStyle4.innerText, 'should have added element to start of head');
-            strictEqual('abc123123', addedStyle4.nonce, 'should have inherited nonce from main script');
+            // Use attribute for compatibility.
+            var nonceAttr = addedStyle4.getAttribute('nonce');
+            nonceAttr = nonceAttr ? nonceAttr : addedStyle4.nonce;
+            strictEqual('abc123123', nonceAttr, 'should have inherited nonce from main script');
             document.head.removeChild(addedStyle4);
 
             // auto escapes JS when through variable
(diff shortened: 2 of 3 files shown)


Template/Tag/CustomHtmlTag.web.js
```diff
@@ -125,6 +125,12 @@
                 newScript.setAttribute('type', element.getAttribute('type'));
             }
 
+            // If this script is executing using a nonce, set the nonce on its children too.
+            var scriptWithNonce = document.querySelector('script[nonce]');
```
tsteur: here we would typically use `TagManager.dom.bySelector` if that works?

tsteur (commented):
Thanks @snake14 Overall that looks but haven't tried to see if that actually works that way or whether the nonce already needs to be defined by the page itself. 

@AltamashShaikh could you maybe try to reproduce this one?

Template/Tag/CustomHtmlTag.web.js
snake14: @tsteur Yes. It looks like that works. I just have to add a little more logic because it returns an array instead of the first element.

snake14:
@tsteur Thank you. In the issue, the user mentioned that they already had a nonce setup in their CSP and that Matomo worked when they included the nonce in the script element in their document header, but it wasn't getting passed on to scripts that Matomo adds to the DOM. So I simply look if the main script element has a nonce and include it when we add custom scripts to the DOM. To reproduce the issue, I added a CSP header, with a nonce, using a .htaccess file. I was able to confirm that the main Matomo/TagManager script worked fine when the nonce attribute was set, but any javascript included in a customHtml Variable wouldn't execute because it was inline and didn't have the nonce. As soon as I made the change in this pull request, it started executing as expected. One option that I was thinking about to make things a little more secure, but would add more responsibility on the user would be to only copy the nonce from the main script if the child script has a custom attribute named something like `data-matomo-include-nonce`.

Template/Tag/CustomHtmlTag.web.js
```diff
@@ -125,6 +125,13 @@
                 newScript.setAttribute('type', element.getAttribute('type'));
             }
 
+            // If this script is executing using a nonce, set the nonce on its children too.
+            var scriptWithNonce = TagManager.dom.bySelector('script[nonce]');
+            if (Array.isArray(scriptWithNonce) && scriptWithNonce.length) {
```
AltamashShaikh: @snake14 This check is failing for me since the type is `object` for me
I removed the `Array.isArray(scriptWithNonce)` check to make the if loop work

Template/Tag/CustomHtmlTag.web.js
```diff
@@ -125,6 +125,13 @@
...
+                scriptWithNonce = scriptWithNonce[0]; // Grab the first element.
+                newScript.nonce = scriptWithNonce.nonce;
```
AltamashShaikh: Also this didn't work for me, I checked on chrome browser,
I did this to make it work `newScript.setAttribute('nonce', scriptWithNonce.getAttribute('nonce'));`

AltamashShaikh: I can see similar behaviour in Firefox too

AltamashShaikh (commented):
Left few comments can you check that @snake14

Template/Tag/CustomHtmlTag.web.js
tsteur: 👍 great. Just fyi we're using these kind of methods for better browser compatibility so that no logic breaks in case the method isn't defined in an older browser.

Template/Tag/CustomHtmlTag.web.js
snake14: That's strange. It uses querySelectorAll which is supposed to return an instance of NodeList, which implements Array. I checked it in both Brave (Chromium) and Firefox and both consoles show as type object, but that it implements the Array prototype. Did you have the nonce attribute set in at least one of your script elements on your test page?

Template/Tag/CustomHtmlTag.web.js
snake14: Interesting. I read that it was better to use the property since the attribute is supposed to be hidden and return and empty string. I tested in Brave (Chromium) and Firefox and it worked correctly in both. I guess I should probably try adding some test cases to see if the tests pass in all of the Travis variations.

Template/Tag/CustomHtmlTag.web.js
AltamashShaikh: Yes I had `nonce` attribute setup on the script element

tests/javascript/index.php
```diff
@@ -2549,6 +2549,10 @@ function callHistory(method)
             fireTemplateTag(templateToTest, params);
             var addedStyle4 = document.getElementById('customStyleTag4');
             strictEqual('var x = {};', addedStyle4.innerText, 'should have added element to start of head');
+            // Use attribute for compatibility.
+            var nonceAttr = addedStyle4.getAttribute('nonce');
+            nonceAttr = nonceAttr ? nonceAttr : addedStyle4.nonce;
```
AltamashShaikh, line 2554: @snake14 was `addedStyle4.getAttribute('nonce')` failing in any case?

AltamashShaikh:
Added 1 small Q rest all looks good :+1:

tests/javascript/index.php
snake14: @AltamashShaikh Yeah. In my local environment, it was failing until I used the property instead of the attribute.

AltamashShaikh: okay

AltamashShaikh (approved):

PR #528, opened 2022-08-01
Variables in Custom JS Funtion Variable

### LINKED ISSUE #212: Support variables in Custom JavaScript variable
# Summary

Using an existing variable into a Custom JavaScript variable can avoid a lot of dev and can be really useful.
Example: you wish return a custom attribute from current clicked element. Use `{{ClickElement}}` is essential.

## Current behavior

Into a custom javascript variable, add the `{{ClickElement}}` variable like :

```javascript
function () { if ({{ClickElement}}) { return {{ClickElement}}.getAttribute('my-attribute'); } else { return ""; } };
```

This code return the following error:

Screen Shot 2020-01-06 at 16 09 16
Screen Shot 2020-01-06 at 16 09 33

Moreother, when this kind of error occured, the Preview mode break totally 😕 

Maybe a similar issue here: https://forum.matomo.org/t/how-to-refer-user-defined-variables-in-custom-javascript/34932

## Expected behavior

Add a variable into a custom javascript variable

ulcuber:
### Description:

Resolves #212

### Review

### DIFF
--- a/Template/Variable/CustomJsFunctionVariable.php
+++ b/Template/Variable/CustomJsFunctionVariable.php
@@ -60,7 +60,7 @@ class CustomJsFunctionVariable extends BaseVariable
                 if (isset($js['joinedVariable']) && is_array($js['joinedVariable'])) {
                     $js = array_reduce($js['joinedVariable'], function ($carry, $item) {
                         if (is_array($item)) {
-                            return $carry . 'TagManager._buildVariable(' . json_encode($item) . ", parameters.get('container'))";
+                            return $carry . 'TagManager._buildVariable(' . json_encode($item) . ", parameters.get('container')).get()";
                         }
                         return $carry . $item;
                     }, "");
--- a/Template/Variable/CustomJsFunctionVariable.php
+++ b/Template/Variable/CustomJsFunctionVariable.php
@@ -36,9 +36,12 @@ class CustomJsFunctionVariable extends BaseVariable
                 $field->title = 'JavaScript Function';
                 $field->description = 'The value should start with "function() { " and end with "return yourValue; }". You have to define a function and return a value. We highly recommend to test the pasted JavaScript function to avoid JavaScript errors on your website.';
                 $field->uiControl = FieldConfig::UI_CONTROL_TEXTAREA;
+                $field->customFieldComponent = self::FIELD_TEXTAREA_VARIABLE_COMPONENT;
+                $field->prepare = function ($value) {
+                    return trim($value);
+                };
                 $field->validators[] = new NotEmpty();
                 $field->validate = function ($value) {
-                    $value = trim($value);
                     if (strpos($value, 'function') !== 0) {
                         throw new \Exception('The value needs to start with "function() { ... }"');
                     }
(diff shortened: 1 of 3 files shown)


AltamashShaikh:
@ulcuber Does this code works for you ?
```
function () { if ({{ClickElement}}) { return {{ClickElement}}.getAttribute('class'); } else { return "111"; } };
```

ulcuber:
@AltamashShaikh checked this issue. It was just creating variable. I forgot to evaluate variable. Now it works

Template/Variable/CustomJsFunctionVariable.php
```diff
@@ -36,30 +36,42 @@ public function getParameters()
                 $field->title = 'JavaScript Function';
                 $field->description = 'The value should start with "function() { " and end with "return yourValue; }". You have to define a function and return a value. We highly recommend to test the pasted JavaScript function to avoid JavaScript errors on your website.';
                 $field->uiControl = FieldConfig::UI_CONTROL_TEXTAREA;
+                $field->customFieldComponent = self::FIELD_TEXTAREA_VARIABLE_COMPONENT;
+                $field->prepare = function ($value) {
```
AltamashShaikh, line 40: @ulcuber any reason why have we added this line ?

AltamashShaikh (commented):
@ulcuber Changes looks good
@tsteur Can you also please review it once ?

Template/Variable/CustomJsFunctionVariable.php
ulcuber: Added because value was trimmed twice

AltamashShaikh: I don't think `prepare` works in anyway, you can simply added the trim back may be ?

tsteur:
@AltamashShaikh @ulcuber the reason I didn't implement it I believe was because I couldn't figure out how to make this work without using `eval`. I believe in some cases this wouldn't work maybe like below?

```js
var {{ foo }} = {{ bar }};
```

Not sure if such cases would work? Or if the variables would only work in few specific places?

Template/Variable/CustomJsFunctionVariable.php
ulcuber: In which case should it not work?

AltamashShaikh: @ulcuber Unable to find any implementation of `prepare` in `https://github.com/matomo-org/matomo/blob/4.x-dev/core/Settings/FieldConfig.php`

ulcuber: https://github.com/matomo-org/matomo/blob/48c458c830fd75ab8cc90489d5a31110ee6c36c9/core/Settings/FieldConfig.php#L213
https://github.com/matomo-org/matomo/blob/48c458c830fd75ab8cc90489d5a31110ee6c36c9/core/Settings/Setting.php#L224

AltamashShaikh: Hey @ulcuber Thanks for pointing it out, my IDE was not recognising the `prepare` changes dunno why.
I can see now :+1:

snake14:
Closing as I believe that this was implemented by #731 . Thank you for your contribution @ulcuber . Sorry that we didn't get it merged.

PR #538, opened 2022-08-12
Fixed start_date and end_date check to fire tags as per UTC

### LINKED ISSUE #511: Improve Start and end date for Tags handling of timezone
We had a user that was having issues when configuring a tag with a specific start date and end date.

While investigating this issue, I found that the Start and End time for a Tag that is stored in UTC will be treated as that time in any timezone and not be converted to that visitors timezone.

For example, if we create a Tag such as the following:
Thursday-14-07-2022 11-24-15
In the above example the start date is set as 1 AM July 14 NZST with end date 8 PM July 14 NZST
When this is converted to UTC, the start date is 1 PM July 13 UTC with end date 8 AM July 14 UTC

When we then test this tag in the browser, the tag does not fire with the following in the logs:
Thursday-14-07-2022 11-24-28
`not firing as this tag does not match date` 
However, the time of my device when testing was 11:20 AM NZST or 11:20PM UTC

When I changed the End Date of the tag to one day further (8 PM July 15 NZST, 8 AM July 15 UTC), it fired correctly as it detected the End Date time had not occurred yet:

Thursday-14-07-2022 11-26-01

This indicates that the time for the start/end date is converted to UTC but is used in the browser as the local timezone of the browser.

AltamashShaikh:
### Description:

Fixed start_date and end_date check to fire tags as per UTC
Fixes: #511.

### Review


tsteur (changes requested):
@AltamashShaikh can we add tests for this?

AltamashShaikh:
> @AltamashShaikh can we add tests for this?

@tsteur Added, please check

javascripts/tagmanager.js
```diff
@@ -425,18 +427,18 @@
 
                     var start, end;
                     try {
-                        start = new Date(startDateTime);
+                        start = new Date(startDateTime + (startDateTime && startDateTime.split(' ').length === 2 ? ' UTC' : ''));
```
tsteur: It took me a while to understand this code and what it does. Maybe we can put this into a function and give a naming so it'll become more clear? Then we could also reuse it in the endDate below.

Part of it why it was hard to understand was because it wasn't clear what `startDateTime` looks like. After looking at tests I figured it must be `2022/01/01 06:00:00` meaning if no timezone is added it will assume the value is in UTC.

javascripts/tagmanager.js
```diff
@@ -425,18 +427,18 @@
...
                         }
                     }
 
                     try {
-                        end = new Date(endDateTime);
+                        end = new Date(endDateTime + (endDateTime && endDateTime.split(' ').length === 2 ? ' UTC' : ''));
                     } catch (e) {
                         if (endDateTime) {
-                            throwError('Invalid startDateTime given');
+                            throwError('Invalid endDateTime given');
```
tsteur, line 441: good find

tsteur (commented):
Nice, that should work @AltamashShaikh  It will improve the situation. Only left a minor comment.

javascripts/tagmanager.js
AltamashShaikh: @tsteur changed to function, I was wondering should we also update the description for start and end date ?
Screenshot from 2022-08-17 12-09-09

Change `If set, this tag will be fired only after this start date. The specified date will be assumed to be in your local timezone.
Your current local time is Wed Aug 17 2022 12:10:31 GMT+0530 (India Standard Time).`

to 
`If set, this tag will be fired only after this start date. The specified date will be converted to UTC datetime for comparison.
Your current local time is Wed Aug 17 2022 12:10:31 GMT+0530 (India Standard Time).`

tsteur: > The specified date will be converted to UTC datetime for comparison.

I wouldn't add it as it might confuse people. They might not know what UTC is and they may be confused why we mention it since they wouldn't need to know that. It's maybe a technical implementation detail?

tsteur: Nice on adding the function 👍  Now we could actually even add a simple test for it? That would also make it easier for other people to understand what the function does when they come across it. A simple basic test or two could do

AltamashShaikh: @tsteur added, please check

tsteur (approved):
👍

PR #544, opened 2022-09-03
When clicking an element within a shadow DOM, use it as the target instead of the shadow root

### LINKED ISSUE #543: Handle Click Events correctly in shadowDOM (web components)
Hi!

I am facing a scenario in which ClickEvents are not handled as expected. 
See also: https://forum.matomo.org/t/event-tracking-inside-webcomponents-shadowdom/47205

After analyzing the scenario I can confirm that when a click occurs on an element (link e.g.) inside the shadowDOM the pushed event occurs on the shadowHost element and not - at least how I would expect from tracking perspective ;) - on the actual clicked element inside the shadow dom.

What is the best way to make it work as intended?

multikoop:
### Description:

This code change enhances the AllElementsClickedTrigger to fire event on the actual clicked target when inside an (open) shadow dom.
Fixes: #543 

### Review

### DIFF
--- a/Template/Trigger/AllElementsClickTrigger.web.js
+++ b/Template/Trigger/AllElementsClickTrigger.web.js
@@ -13,6 +13,11 @@
                 return;
             }
             var target = event.target;
+
+            if (target.shadowRoot) {
+                target = event.composedPath()[0];
+            }
+
             triggerEvent({
                 event: 'mtm.AllElementsClick',
                 'mtm.clickElement': target,
--- a/Template/Trigger/AllElementsClickTrigger.web.js
+++ b/Template/Trigger/AllElementsClickTrigger.web.js
@@ -12,10 +12,13 @@
             if (!event.target) {
                 return;
             }
-            var target = event.target;
 
+            var target = event.target;
             if (target.shadowRoot) {
-                target = event.composedPath()[0];
+                var composedPath = event.composedPath();
+                if (composedPath.length) {
+                      target = composedPath[0];   //In shadow DOM select the first event path as the target
+                }
             }
 
             triggerEvent({
--- a/Template/Trigger/AllDownloadsClickTrigger.web.js
+++ b/Template/Trigger/AllDownloadsClickTrigger.web.js
@@ -21,11 +21,14 @@
                         return;
                     }
 
-                    var target = event.target;
+                    var target = event.target;                    
                     if (target.shadowRoot) {
-                        target = event.composedPath()[0];
+                        var composedPath = event.composedPath();
+                        if (composedPath.length) {
+                            target = composedPath[0];   //In shadow DOM select the first event path as the target
+                        }
                     }
-                    
+
                     var nodeName = target.nodeName;
 
                     while (!isClickNode(nodeName) && target && target.parentNode) {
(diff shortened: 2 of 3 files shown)


MatomoForumNotifications:

https://forum.matomo.org/t/event-tracking-inside-webcomponents-shadowdom/47205/2

snake14 (approved):
Successfully tested this change and it looks good to me. Any concerns @AltamashShaikh ?

AltamashShaikh:
@snake14 Changes look good, but we need to add this in other click triggers too `AllDownloadsClickTrigger.web.js`, `AllLinksClickTrigger.web.js`,

justinvelluppillai:
Is it going to work in older browsers?

snake14:
@justinvelluppillai That's a good point. According to [Can I use](https://caniuse.com/?search=shadowRoot), it's supported by pretty much all newer browsers. So, it doesn't look like it would work for IE or some older browsers. Is that a deal breaker?

justinvelluppillai:
This line of code would just need to not run and not break or error for older browsers, so just making sure it's adequately handling that is the key.

snake14:
@justinvelluppillai I would expect `target.shadowRoot` to evaluate as undefined for older browsers. So, I think it would be alright.

multikoop:
Thanks for the discussions. Yes I also was thinking of an enhancement for other relevant triggers (like mentioned by @AltamashShaikh ) Since the AllDownloadsClickTrigger the clickCallBack  has advanced implementation  I do not feel experienced enough with the needed code change. Maybe someone else.... :-)

multikoop:
As mentioned by @AltamashShaikh I have added the code change (to deal with the target element inside shadow dom) to `AllDownloadsClickTrigger.web.js` and  `AllLinksClickTrigger.web.js`. so that the change is consistent in available All*ClickTriggers

multikoop:
Hhhm, any idea why 1 build job failed?

snake14:
@multikoop Yep. It looks like there was a  random server error during one of the specific builds. I restarted it and it ran successfully this time. It's all good now.

Template/Trigger/AllElementsClickTrigger.web.js
```diff
@@ -13,6 +13,11 @@
                 return;
             }
             var target = event.target;
+
+            if (target.shadowRoot) {
+                target = event.composedPath()[0];
```
AltamashShaikh: @multikoop Can we update it to below code and add a comment so that in future we remember why we need to select the first element for the event array
```suggestion
                var composedPath = event.composedPath();
                if (composedPath.length) {
                      target = composedPath[0];   //In shadow DOM select the first event path as the target
                }
```

AltamashShaikh (commented):
@multikoop Left 1 minor comment to make code more readable

Template/Trigger/AllElementsClickTrigger.web.js
multikoop: Thanks @AltamashShaikh  I have added the code change to all All*ClickTrigger.web.js.

AltamashShaikh (approved):

AltamashShaikh:
@multikoop Thank you for your contribution, your changes should go live with release of Matomo 4.12

multikoop:
Thanks for the discussion and accepting the contribution. Looking forward to the release 🤩

PR #563, opened 2022-11-01
Add custom visitor cookie timeout setting

tomper00:
Add possibility to set custom a custom cookie timeout in the Matomo config variable as described in https://github.com/matomo-org/tag-manager/issues/562

### Description:

This will add a new setting in the Matomo Config variable where you can manually set the timeout in seconds for the pk_id cookie. 
 

### Review

* [ ] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
* [ ] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [ ] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [ ] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)
* [ ] [Code review done](https://developer.matomo.org/guides/pull-request-reviews#code-review-done)
* [ ] [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)
* [ ] [Reviewed for breaking changes](https://developer.matomo.org/guides/pull-request-reviews#reviewed-for-breaking-changes)
* [ ] [Developer changelog updated if needed](https://developer.matomo.org/guides/pull-request-reviews#developer-changelog-updated-if-needed)
* [ ] [Documentation added if needed](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)
* [ ] Existing documentation updated if needed

PR #566, opened 2022-11-04
Add a new field custom URL Referrer on pageview tag

tomper00:
### Description:

This is often needed when you work with SPA tracking since the Referrer often is not updated on virtual pageviews.
It is also nice to be able to create a custom JS Variable some times to be able to filter out PII etc. 

### Review


snake14 (changes requested):
Thank you for taking the time to create this PR, @tomper00 . I think it looks really good. The only request I have is that you create a migration similar to [the one you created in your other PR](https://github.com/matomo-org/tag-manager/pull/563/files#diff-28ece140c822d7c3d1db0e3bafb1d05e716b457a1b4e10e5912794378b532e82). I guess another option would be to wait until your other PR is merged, merge those changes into this branch, and then simply add a single `$migrator->addField('customUrlRef');` line to the migration your created as part of your other PR.

Oh. One other side note. Please remember to add the `Needs Review` label to any PRs that you create in the future. It just makes it more visible to our team and lets us know that it's ready for us to provide feedback.

tomper00:
Ok thanks again for the feedback @snake14 I will go with the option to merge my branches, that sounds like an easy way forward.

snake14:
> Ok thanks again for the feedback @snake14 I will go with the option to merge my branches, that sounds like an easy way forward.

@tomper00 I just left a comment on your other PR, but I realised that I had you use the wrong class for the migration. Since this one is for a tag, you would use the `NewTagParameterMigrator` class. You can still merge your branches, but you'd need two lines instead of just the one:
```
$migrator = new NewTagParameterMigrator(MatomoTag::ID, 'customUrlRef');
$migrator->migrate();
```
So, if you merge the two branches, you would run `NewTagParameterMigrator->migrate()` for your tag column and `NewVariableParameterMigrator->migrate()` for your two variable columns.

lang/en.json
```diff
@@ -140,6 +140,8 @@
         "CustomTitleHelp": "Optionally, specify a custom document title which should be tracked instead of the default document title.",
         "CustomUrl": "Custom URL",
         "CustomUrlHelp": "Optionally, specify a custom URL which should be tracked instead of the current location.",
+        "CustomUrlRef": "Custom URL Referrer",
+        "CustomUrlRefHelp": "Optionally, specify a custom URL referrer which should be tracked instead of the current urlref. This is handy when you are working with SPA applications since the url ref often is not updated correctly. Another scenario would be if you want to do some kind of replacing or modifications to it before its sent to Matomo.",
```
AltamashShaikh: @tomper00 Does the below suggestion makes more sense to the user ?
```suggestion
        "CustomUrlRefHelp": "Optionally, specify a custom URL referrer which should be tracked instead of the current URL referrer. This is handy when you are working with SPA applications since the URL referrer is not updated correctly. Another scenario would be if you want to do some kind of replacement or modifications to the referrer URL to be tracked by Matomo.",
 ```

File: Updates/4.12.4-b2.php [hunk 2]
```diff
@@ -0,0 +1,28 @@
...
+ * Update for version 4.12.4-b1.
+ */
+class Updates_4_12_4_b3 extends PiwikUpdates
+{
+    public function doUpdate(Updater $updater)
+    {
+        $migrator = new NewTagParameterMigrator(MatomoTag::ID, 'customUrlRef');
+        $migrator->migrate();
+    }
+}
```
snake14: Since you merged the changes from your other branch into this one, you don't have to create a new migration script. You should be able to add the two lines from the `doUpdate` method in this one to the end of the `doUpdate` in [the existing migration script](https://github.com/matomo-org/tag-manager/commit/1c52824eb4c0be4c35369376c31c22aea83f4f06#diff-28ece140c822d7c3d1db0e3bafb1d05e716b457a1b4e10e5912794378b532e82). That way both your variable and tag changes get migrated in the same script. I guess it doesn't hurt to have two separate scripts, but it makes the release process a little more complicated.

snake14 (dismissed):
Looks good. I just left one comment.

Updates/4.12.4-b2.php
AltamashShaikh: @snake14 @tomper00 You should move this changes to `5.x-dev` branch since the next release is `5.x-dev`

Updates/5.0.0-b2.php
```diff
@@ -0,0 +1,28 @@
...
+use Piwik\Plugins\TagManager\Template\Tag\MatomoTag;
+use Piwik\Plugins\TagManager\Template\Variable\MatomoConfigurationVariable;
+use Piwik\Plugins\TagManager\Updates\NewVariableParameterMigrator;
+use Piwik\Updater;
+use Piwik\Updates as PiwikUpdates;
+
+/**
+ * Update for version 4.12.4-b1.
+ */
+class Updates_4_12_4_b3 extends PiwikUpdates
```
snake14, line 21: I believe that this will error out if the class name doesn't match the file name. Could you please update it?

snake14: @tomper00 would you please be able to update the class name to match the file name?

AltamashShaikh:
@tomper00 Any update on this ?

AltamashShaikh:
We already added this functionality, so closing this PR as its already done.

image

PR #594, opened 2023-01-20
Improvements to MTM tracker to allow full _paq functionality

snake14:
### Description:

This is to remedy the fact that _paq.push doesn't work reliably while using MTM.

### Review


snake14:
@AltamashShaikh Any ideas what is causing the JavaScript build to fail? It appears to have started with the [previous commit](https://github.com/matomo-org/tag-manager/pull/594/commits/9fb7e808e479bf1501138186639eaabadee56e56), but I'm not seeing what in that commit would cause the failure.

AltamashShaikh:
> @AltamashShaikh Any ideas what is causing the JavaScript build to fail? It appears to have started with the [previous commit](https://github.com/matomo-org/tag-manager/pull/594/commits/9fb7e808e479bf1501138186639eaabadee56e56), but I'm not seeing what in that commit would cause the failure.

Seeing the same issue in FormAnalytics build and trying to figure out, @sgiehl do you have any idea why this is breaking ?

AltamashShaikh (commented):
@snake14 looks good, but why haven't we made any changes for `ecommerce` to work ?
Also can you add an example html [here](https://github.com/matomo-org/tag-manager/tree/4.x-dev/tests/resources) to test this functionality better ?

snake14:
> @snake14 looks good, but why haven't we made any changes for `ecommerce` to work ? Also can you add an example html [here](https://github.com/matomo-org/tag-manager/tree/4.x-dev/tests/resources) to test this functionality better ?

@AltamashShaikh I didn't think that any specific changes were necessary to make `ecommerce` work because it's using the default Tracker from `js/piwik.js`. Did you test tracking `ecommerce` using `_paq.push`?

Template/Tag/MatomoTag.web.js
```diff
@@ -14,6 +14,13 @@
     };
 
     window._paq = window._paq || [];
+    // Store the initial state of window._paq so that we can apply it to all of the configs
+    var initialPaq = window._paq && window._paq.length ? JSON.parse(JSON.stringify(window._paq)) : initialPaq || [];
```
AltamashShaikh, line 19: @snake14 Can you add a comment why do we need to stringify and convert it back to JSON here ?

javascripts/tagmanager.js
```diff
@@ -1673,8 +1673,26 @@
                 }
 
                 if (container.tags && utils.isArray(container.tags)) {
```
AltamashShaikh, line 1675: @snake14 Can you also add a comment what are we doing here, its a bit hard to understand

Template/Tag/MatomoTag.web.js
```diff
@@ -117,39 +128,167 @@
                     }
                     configuredTrackers[variableName] = tracker;
 
+                    var requireCookieConsentIndex = disableBrowserFeatureDetectionIndex
```
AltamashShaikh: Also can we document this, whenever any new config is supported we should not forget to add new variable here and support it.

AltamashShaikh (commented):
@snake14 Left few suggestion for comments and doc and was able to test with the resources you shared and looks good to me

Template/Tag/MatomoTag.web.js
snake14: Sure. I'm adding a comment now.

javascripts/tagmanager.js
snake14: I didn't add line 1675, but I added some more inline comments in my new lines below.

Template/Tag/MatomoTag.web.js
snake14: Sure. Do you think it would be best to note it in the developer documentation, the FAQ, or both? Any specific page recommendations?

AltamashShaikh: Developer documentation should be fine

AltamashShaikh (approved):
Except for adding the documentation all looks good :+1:

snake14:
Thank you @AltamashShaikh . I went ahead and created a [documentation PR](https://github.com/matomo-org/developer-documentation/pull/700), so I'm going to merge this PR.

sgiehl:
@snake14 @AltamashShaikh It doesn't look like a useful move to simply merge a PR that is known to break the tests.
This might be fine if tests are failing randomly or it's known the failure is caused by something else. But in this case it's clear the javascript tests started to fail with https://github.com/matomo-org/tag-manager/pull/594/commits/9fb7e808e479bf1501138186639eaabadee56e56
So some changes in that commit are making problems. I'm not deep enough in the javascript of the TagManager, but I would assume that resetting `window._paq` might cause the trouble. If you need help with solving that, maybe @tsteur can assist.

snake14:
@sgiehl We weren't sure since the errors were identical to the JavaScript build errors in other plugin branches. Plus, the changes have been tested by Altamash, Thomas, and myself without issue. Is the issue in the code or is the build too brittle? I can try a slightly different approach around `window._paq`, but I don't know if that's the root issue.

Edit: I created [a new PR](https://github.com/matomo-org/tag-manager/pull/599) and tried a few changes to make sure that I wasn't reassigning `window._paq` or `_paq`. It looks like the build is passing now. Thank you for the suggestion @sgiehl . I'm still seeing some 404 errors in the build's output, but I need to move on from this issue.

AltamashShaikh:
@sgiehl Thanks for pointing it out and we will be careful next time, actually I did test this on local and it seems to passing so we thought it as a recent random JS tests failing on github-actions and proceeded ahead

sgiehl:
@AltamashShaikh If you think something fails randomly simply try to rerun the job a couple of times and see if its really random or fails always...

PR #656, opened 2023-07-04
Improved steps for React no data screen tab

AltamashShaikh:
### Description:

Improved steps for React no data screen tab
Fixes: #PG-2882

### Review


michalkleiner:
My instinct is that at some point we will get bitten by the fact we are reusing the same template and translation strings in two separate no-data screen tabs even though they are the same apart from step 16. It just doesn't feel right, but I guess it technically works.

How likely it is the user will see both at the same time?

AltamashShaikh:
> My instinct is that at some point we will get bitten by the fact we are reusing the same template and translation strings in two separate no-data screen tabs even though they are the same apart from step 16. It just doesn't feel right, but I guess it technically works.
> 
> How likely it is the user will see both at the same time?

@michalkleiner It was done to avoid duplication, I hope it doesn't bite us back in future
Also regarding your other Question, I don't have an answer that how many will view both at same time, looks unlikely

snake14 (approved):
Looks good. The tab displays as expected when I test locally :+1:

PR #674, opened 2023-08-17
Add tracking code install test to no data tab

snake14:
### Description:

This depends on the core PR: [21148](https://github.com/matomo-org/matomo/pull/21148)
We're adding the tracking code install check to the no data tab for MTM.

### Review


lang/en.json
```diff
@@ -913,6 +913,9 @@
...
-        "ReactFollowStep16": "Inject the %1$sMatomo Tag Manager JS%2$s code into your %3$s (or other relevant files), we recommend doing this using the \"%4$s\" method. The example below shows how to add the %1$sMatomo Tag Manager JS%2$s code into a \"%5$s\" app in %6$s."
+        "ReactFollowStep16": "Inject the %1$sMatomo Tag Manager JS%2$s code into your %3$s (or other relevant files), we recommend doing this using the \"%4$s\" method. The example below shows how to add the %1$sMatomo Tag Manager JS%2$s code into a \"%5$s\" app in %6$s.",
+        "SiteWithoutDataMtmIntro": "The Matomo Tag Manager may be a better choice than tracking via the regular JavaScript code. It lets you manage and unify all your tracking and marketing tags (also known as \"pixels\" or \"custom HTML/JS tags\") in one place. Using the Matomo Tag Manager makes your life easier when you want to modify any of these tags, as you will no longer need a developer to make the needed changes for you. %1$sLearn more%2$s.",
+        "SiteWithoutDataMtmStep2": "If you would like to customize this container, visit the ",
```
sgiehl: Looking at the other code it seems the sentence is follow by `Container Dashboard`, which is currently added by using translations for those single strings. It might be better to include both words here directly. Even though the words might need to be translated here as well, simply appending them might be incorrect in various languages. als the `the` in the end would be hard to translate for translators without knowing what follows...

snake14: Thank you @sgiehl . I thinking of maintainability if the Container Dashboard name was ever changed, but you make a good point about sentence structure in other languages and context. I'll add `Container Dashboard` into the translation string.

AltamashShaikh (approved):
@snake14 I tested and changes looks good, I notice that if my measurable has no URL added it tries to open the Matomo instance any says tracking code not installed error. Do you think we should rather alert the user to map the URL ?

Also check with Javi/Stan if they think the note should be indented a bit to right like below 
Screenshot from 2023-08-23 07-56-52

snake14:
> @snake14 I tested and changes looks good, I notice that if my measurable has no URL added it tries to open the Matomo instance any says tracking code not installed error. Do you think we should rather alert the user to map the URL ?
> 
> Also check with Javi/Stan if they think the note should be indented a bit to right like below

Thank you @AltamashShaikh . I'll check with Product about the desired behaviour when a URL hasn't been configured for a site. I checked about the alignment of the success message and they said it's correct how it is.

michalkleiner:
Tested the core part and that works as expected. Looked the MTM PR here and when there's no container, the logic seems to be broken. It does say there's no published version, but the text for step 1 is missing and the installation test probably shouldn't be available at that point. Good think is that it still successfully fails.

Screenshot 2023-09-06 at 5 55 58 PM

snake14:
> Tested the core part and that works as expected. Looked the MTM PR here and when there's no container, the logic seems to be broken. It does say there's no published version, but the text for step 1 is missing and the installation test probably shouldn't be available at that point. Good think is that it still successfully fails.

Thank you @michalkleiner . I thought I had that working, but it might have been changed while I was making revisions. I'll look into it.

AltamashShaikh (changes requested):
@snake14 I tested and the changes look good, now since we have made the URL configurable. user can enter anything and I see we do have a check for valid URL, but if we add a invalid URL we get a error but loading icon keeps showing up with the input disabled.

Screenshot from 2023-09-07 08-02-06

snake14:
> @snake14 I tested and the changes look good, now since we have made the URL configurable. user can enter anything and I see we do have a check for valid URL, but if we add a invalid URL we get a error but loading icon keeps showing up with the input disabled.

That's a good catch @AltamashShaikh . I believe that core team added that validation before merging the core changes. I wonder if a core issue should be created as that issue can't be fixed in the plugin.

AltamashShaikh:
> > @snake14 I tested and the changes look good, now since we have made the URL configurable. user can enter anything and I see we do have a check for valid URL, but if we add a invalid URL we get a error but loading icon keeps showing up with the input disabled.
> 
> That's a good catch @AltamashShaikh . I believe that core team added that validation before merging the core changes. I wonder if a core issue should be created as that issue can't be fixed in the plugin.

@snake14 Yes create a core issue to fix it, and this will be fixed automatically :+1:

sgiehl:
@AltamashShaikh I've created a PR in core to fix the handling of invalid URLs. 
Is this TagManager PR ready to merge? I'd like to release a new RC of Matomo today, and might be good to include this changes if possible.

AltamashShaikh:
@sgiehl It good to be merged, will merge it

AltamashShaikh (approved):

Issue #666, opened 2023-08-02
HeartBeat timer in matomo tag manager


Chardonneaur:
Hello,
would it be possible to set the duration of the timer within the heartbeat timer feature of Matomo Tag Manager?

snake14:
Hi @Chardonneaur . Looking at the code, it appears that Tag Manager only uses the default setting for the heartbeat timer, which I believe is 15 seconds. You could probably use `_paq.push(['enableHeartBeatTimer', {desired_number_over_5}]);` inside of a CustomHtml tag. Will that work or would you like me to add this for prioritisation as an enhancement?

Chardonneaur:
Thank you @snake14, well i think that could be a great enhancement for any features of the tag manager where a value can be entered as an additional option.

snake14:
Sounds good. I'm marking this as an enhancement to be prioritised.

Just to add more detail to help with the prioritisation process, we're talking about adding an optional field where the customer can set the number of seconds when the Enable HeartBeat Timer checkbox is checked in a Matomo configuration Variable:
![image](https://github.com/matomo-org/tag-manager/assets/11394099/9f68d8af-e9f9-4942-8770-050d954d0767)

Chardonneaur:
Yes exactly, the idea would be to have the same level of customization within the Matomo configuration variable than on a custom tracking code.

EPNW-Eric:
Anyone working on this yet? If not, I'd like to try implementing it 😃.

AltamashShaikh:
@EPNW-Eric Feel free to work on this :+1:

heurteph-ei:
I thought the HeartBeat was no more a ping request, but has been replaced by the sendbeacon at page unload... Am I wrong?

AltamashShaikh:
@heurteph-ei Thanks for the info, I just checked with the core team and what you are saying seems to be correct.
https://github.com/matomo-org/matomo/issues/15612#issuecomment-664615366
@EPNW-Eric Can we please remove the configuration from the PR ?

EPNW-Eric:
Hm, even if @heurteph-ei is correct with respect to the `unload`-event, the `configHeartBeatDelay` is still used in calculations to determin wheter to send a ping or not:

https://github.com/matomo-org/matomo/blob/d58c2ab0328ff9189e43cdd1ea31559e0d563d90/js/piwik.js#L2951-L2957

[...]

PR #679, opened 2023-08-22
Several new options for the Matomo Configuration

EPNW-Eric:
The current Matomo Tracker Configuration UI lack some of the things that can be done using JavaScript. I added some of the missing options to the Tag Manager. Some of them were needed by us for a customer (e.g. `requestMethod`, `trackingEndpointCustom`), while some others just seemed easy to implement (e.g. `enableFileTracking`, `heartbeatTime`).

I haven't done heavy testing yet nor did I write any tests. The reasoning behind that is, that I don't want to invest that time in features that you don't want to include in the Tag Manager. So before more testing is done, we should discussions about the features.

Here are the features:
- if `bundleTracker` is checked, `jsEndpoint` option gets hidden, placed `jsEndpoint` option next to `bundleTracker`
- added `heartBeatTime` (#666)
- ~~changed `customCookieTimeOutEnable` title and description (english only yet)~~
- ~~renamed `customCookieTimeOut` to `visitorCookieTimeOut`~~
- added `referralCookieTimeOut`
- added `sessionCookieTimeOut`
- added `enableFileTracking`
- added `crossDomainLinkingTimeout`
- added `cookieNamePrefix`
- added `disablePerformanceTracking`
- added `trackingEndpointCustom` and extended `trackingEndpoint` (partially matomo-org/matomo#19864)
- added `jsEndpointCustom` and extended `jsEndpoint`
- added `appendToTrackingUrl`
- added `forceRequestMethod`
- added `requestMethod`
- added `requestContentType`
- added `CustomRequestProcessingVariable`, `CustomRequestProcessing` (as `Validator`) and `customRequestProcessing` (for the tracker)

I think the last point, `CustomRequestProcessing` stands out a bit. I added a new `Variable` type and a new `Validator`. For choosing a variable of type `CustomRequestProcessingVariable` in the UI I at first wanted to use the same mechanism as for selecting a `MatomoConfigurationVariable` in the `MatomoTag`
```php
$field->uiControlAttributes = array('variableType' => 'MatomoConfiguration');
```
so I wouldn't have needed the `Validator`, but it turned out that there is currently no way to reset something that was selected with the `variableType`-mechanism.

I look forward to discussing the features and hope to see many of them making it into the Tag Manager, so we can get closer to feature parity 😃.


snake14:
Thank you for taking the time to put together this PR @EPNW-Eric . I haven't done any functional testing, but I think the code looks pretty good. Do you see any concerns @AltamashShaikh ?

lang/en.json
```diff
@@ -559,16 +570,22 @@
         "MatomoConfigurationMatomoDisableCookiesDescription": "Disables all first party cookies.",
         "MatomoConfigurationMatomoRequireConsentTitle": "Require tracking consent",
         "MatomoConfigurationMatomoRequireConsentDescription": "Track only when user gave tracking consent. In a consent screen you need to call \"window._paq=window._paq||[];window._paq.push(['rememberConsentGiven']);\" when the user gives consent.",
-        "MatomoConfigurationMatomoCustomCookieTimeOutEnableTitle": "Enable custom visitor Cookie timeout",
-        "MatomoConfigurationMatomoCustomCookieTimeOutEnableDescription": "Allows you to set a custom time in full days for the _pk_id cookie.",
-        "MatomoConfigurationMatomoCustomCookieTimeOutTitle": "The pk_id expire time in full days",
-        "MatomoConfigurationMatomoCustomCookieTimeOutDescription": "Timeout in full days. By default, the Matomo tracking cookie expires in 13 months (365 + 28 days) = 393 days.",
+        "MatomoConfigurationMatomoCustomCookieTimeOutEnableTitle": "Enable Custom Cookie Timeouts",
```
AltamashShaikh: @EPNW-Eric Since this keys are already translated, you should create a new one rather than updating the current one, something like below or any name you think is correct
```suggestion
        "MatomoConfigurationMatomoCustomVisitorCookieTimeOutEnableTitle": "Enable Custom Cookie Timeouts",
```

Template/Variable/MatomoConfigurationVariable.php
```diff
@@ -141,10 +163,22 @@ public function getParameters()
                 $field->description = Piwik::translate('TagManager_MatomoConfigurationMatomoCustomCookieTimeOutEnableDescription');
                 $field->condition = '!disableCookies';
             }),
-            $this->makeSetting('customCookieTimeOut', '393', FieldConfig::TYPE_INT, function (FieldConfig $field) {
```
AltamashShaikh, line 144: @EPNW-Eric will it be okay if we do not change the name here ?

AltamashShaikh (commented):
@EPNW-Eric Took a glance and left few comments, but this looks really good :) 
I am yet to do any functional testing, but I think you should progress with the PR.

Also note you need to add a migration script for new fields - see this [file](https://github.com/matomo-org/tag-manager/blob/5.x-dev/Updates/5.0.0-b1.php) for reference

Template/Variable/MatomoConfigurationVariable.php
EPNW-Eric: Hm I think we could keep the name `customCookieTimeOut`, but in my opinion it would be cleared if its changed to `visitorCookieTimeOut` since it's just the timeout for the visitor cookie, but not for the other two types of cookies (session and referral). Also, `visitorCookieTimeOut` would be consistent with the new `sessionCookieTimeOut` and `referralCookieTimeOut`. Is the reason behind keeping it migration?

lang/en.json
EPNW-Eric: The key is just used once, so changing the content would not affect any other places in the UI. This also means that if we add a new key, this old key is never used anywhere again and just obsolete. What if we keep the update-approach but I delete the key from all other language files?

If a new-key-approach is preferable to this, I think `MatomoConfigurationMatomoCustomCookieTimeOutsEnableTitle` would fit, even if it just differs from the old key by the plurals *s* in *TimeOuts*. Would that be fine or is it too confusing?

EPNW-Eric:
> @EPNW-Eric Took a glance and left few comments, but this looks really good :) I am yet to do any functional testing, but I think you should progress with the PR.
> 
> Also note you need to add a migration script for new fields - see this [file](https://github.com/matomo-org/tag-manager/blob/5.x-dev/Updates/5.0.0-b1.php) for reference

Sure 😃 Should I append to the old file `5.0.0-b1.php` or create a new one? If so, what should it be named?

lang/en.json
AltamashShaikh: if you search for `MatomoConfigurationMatomoCustomCookieTimeOutEnableTitle` you would notice that its already translated in other languages,  and it won't get translated so that's why we are asking you to use new keys and yes you can delete the old text from `en.josn` if it is not being used anywhere else

Template/Variable/MatomoConfigurationVariable.php
AltamashShaikh: @EPNW-Eric Yes the reason is to keep the migration simple, you can adjust the description to mention that it will affect the visitor cookie timeout only

AltamashShaikh:
> 5.0.0-b1.php

@EPNW-Eric We are already on `rc-1` for Matomo core, create a new file `5.0.0-rc2.php`

EPNW-Eric:
I don't get the tests to pass and I don't know why...

snake14:
@EPNW-Eric The code changes look good. I'm not quite sure why the tests are failing, but I haven't taken a close look.

Template/Variable/MatomoConfigurationVariable.php
```diff
@@ -283,9 +325,19 @@ public function getParameters()
                     'piwik.js' => 'piwik.js',
                     'js/' => 'js/',
                     'js/tracker.php' => 'js/tracker.php',
+                    '<custom>' =>  Piwik::translate('TagManager_MatomoConfigurationMatomoJsEndpointCustom'),
```
AltamashShaikh: @EPNW-Eric I updated all the `<custom>` with `custom` and reverted the changes added in tests files and ran the tests again and that seems to work

Screenshot from 2023-08-27 14-01-13

AltamashShaikh (commented):
@EPNW-Eric Left a comment om how to fix the test cases

EPNW-Eric:
> @EPNW-Eric Left a comment om how to fix the test cases

Thank you, I'll look into it soon, I'm a bit busy at the moment...

EPNW-Eric:
I rebased the commits onto the current master and fixed the tests. The UI tests are still failing (I guess because of the screenshots?).

snake14:
Thank you @EPNW-Eric . We can take care of the UI build after it's merged. @AltamashShaikh should we be concerned about the JavaScript build failures?

AltamashShaikh:
@snake14 @EPNW-Eric The failing JS tests seems to be failing for this branch only, so we should fix it..UI tests can be handled by us post merge

EPNW-Eric:
I can look into JS more, but I'm currently not in the office. It will take like 2 weeks for me until I can investigate it further, so don't panic thinking this PR got stale if you don't hear from me 😅

EPNW-Eric:
So I checked the code again and the first thing I noticed was a missing `}`. I added it and hope this fixes the errors... Can you approve the test workflow?

EPNW-Eric:
> So I checked the code again and the first thing I noticed was a missing `}`. I added it and hope this fixes the errors... Can you approve the test workflow?

This seems to have fixed the JS tests, the other tests started failing, I think because of a formatting issue. The latest push should have fixed that.

Template/Variable/CustomRequestProcessingVariable.php
```diff
@@ -0,0 +1,69 @@
...
+        return array(
+            $this->makeSetting('jsFunction', "function(request){\n\treturn request;\n}", FieldConfig::TYPE_STRING, function (FieldConfig $field) {
+                $field->title = Piwik::translate('TagManager_CustomRequestProcessingVariableJsFunctionTitle');
+                $field->description = Piwik::translate('TagManager_CustomRequestProcessingVariableJsFunctionDescription');
+                $field->uiControl = FieldConfig::UI_CONTROL_TEXTAREA;
+                $field->validators[] = new NotEmpty();
+                $field->validate = function ($value) {
+                    $value = trim($value);
+                    if (strpos($value, 'function(request){') !== 0) {
+                        throw new \Exception('The value needs to start with "function(request){"');
```
snake14, line 46: Since this is a validation error that will be surfaced to the customer, I wonder if it should be translated?

snake14 (approved):
Nice work! I just left one comment.

Template/Variable/CustomRequestProcessingVariable.php
```diff
@@ -0,0 +1,69 @@
...
+
+    public function isCustomTemplate()
+    {
+        return true;
+    }
+
+    public function getParameters()
+    {
+        return array(
+            $this->makeSetting('jsFunction', "function(request){\n\treturn request;\n}", FieldConfig::TYPE_STRING, function (FieldConfig $field) {
```
AltamashShaikh: @EPNW-Eric Do we need this `\n` here?

Template/Variable/CustomRequestProcessingVariable.php
```diff
@@ -0,0 +1,69 @@
...
+                $field->title = Piwik::translate('TagManager_CustomRequestProcessingVariableJsFunctionTitle');
+                $field->description = Piwik::translate('TagManager_CustomRequestProcessingVariableJsFunctionDescription');
+                $field->uiControl = FieldConfig::UI_CONTROL_TEXTAREA;
+                $field->validators[] = new NotEmpty();
+                $field->validate = function ($value) {
+                    $value = trim($value);
+                    if (strpos($value, 'function(request){') !== 0) {
```
AltamashShaikh, line 45: should we use `stripos` here to make the comparison case insensitive?

Updates/5.0.0-rc2.php
```diff
@@ -0,0 +1,80 @@
...
+use Piwik\Plugins\TagManager\UpdateHelper\NewVariableParameterMigrator;
+use Piwik\Updater;
+use Piwik\Updater\Migration;
+use Piwik\Updater\Migration\Factory as MigrationFactory;
+use Piwik\Updates as PiwikUpdates;
+
+/**
+ * Update for version 5.0.0-rc2.
+ */
+class Updates_5_0_0_rc2 extends PiwikUpdates
```
AltamashShaikh: @EPNW-Eric [5.0.0-rc2](https://github.com/matomo-org/matomo/releases/tag/5.0.0-rc2) was released 4 days ago can you change it to rc3 ?

AltamashShaikh (changes requested):
No FAQ link, and the description is confusing like for GET it says POST
Screenshot from 2023-09-11 08-35-02

Template/Variable/CustomRequestProcessingVariable.php
EPNW-Eric: We could, but I wanted it consistent with the existing code here: https://github.com/matomo-org/tag-manager/blob/e3c5a5ecd5b1e968f2bba98eaf1e19909ac78ebf/Template/Variable/CustomJsFunctionVariable.php#L43

so i also used `strpos`. I have no problem changing it to `stripos`, but then I'd like to change it in `CustomJsFunctionVariable.php`, too. Just let me know what you prefer 😄

Template/Variable/CustomRequestProcessingVariable.php
EPNW-Eric: I wanted it to be consistent with the existing code here: https://github.com/matomo-org/tag-manager/blob/e3c5a5ecd5b1e968f2bba98eaf1e19909ac78ebf/Template/Variable/CustomJsFunctionVariable.php#L44

so i also did not use translation. I have no problem using translation, but then I'd like to change it in `CustomJsFunctionVariable.php`, too. Just let me know what you prefer 😄

Template/Variable/CustomRequestProcessingVariable.php
AltamashShaikh: @EPNW-Eric Yes good to add it there too :+1:

Template/Variable/CustomRequestProcessingVariable.php
EPNW-Eric: Nope, that was just personal preference I guess...
But you are correct, the exisiting code in https://github.com/matomo-org/tag-manager/blob/e3c5a5ecd5b1e968f2bba98eaf1e19909ac78ebf/Template/Variable/CustomJsFunctionVariable.php#L36
 also does not use `\n`, so I will remove it here.

Updates/5.0.0-rc2.php
EPNW-Eric: Will do 👍

EPNW-Eric:
> No FAQ link, and the description is confusing like for GET it says POST Screenshot from 2023-09-11 08-35-02

You are correct, the link is missing, I will add that.
As for the confusing desciprtion, I just copied it from the JavaScript Tracker documentation:
https://github.com/matomo-org/developer-documentation/blob/75999a432b31f0ecf445162fe3d810c2cd651f8c/docs/5.x/tracking-javascript.md?plain=1#L158

Even if it's a bit confusing I'd prefer to keep both documentations the same since in my opinion different documentations for the same thing always cause more trouble... Let me know what you think 😃

AltamashShaikh:
> > No FAQ link, and the description is confusing like for GET it says POST Screenshot from 2023-09-11 08-35-02
> 
> You are correct, the link is missing, I will add that. As for the confusing desciprtion, I just copied it from the JavaScript Tracker documentation: https://github.com/matomo-org/developer-documentation/blob/75999a432b31f0ecf445162fe3d810c2cd651f8c/docs/5.x/tracking-javascript.md?plain=1#L158
> 
> Even if it's a bit confusing I'd prefer to keep both documentations the same since in my opinion different documentations for the same thing always cause more trouble... Let me know what you think 😃

@EPNW-Eric Just add the FAQ link :+1:

Template/Variable/CustomRequestProcessingVariable.php
EPNW-Eric: Wait a second, in JavaScript, `function` is a keyword so it has to be lower-case 😅, so using `strpos` is correct... We should not change it.

Template/Variable/CustomRequestProcessingVariable.php
EPNW-Eric: So I double checked an noticed that exception translation is not consistent.

Regarding Variables, they are (in addition to `CustomJsFunctionVariable.php` as above) not translated in
https://github.com/matomo-org/tag-manager/blob/e3c5a5ecd5b1e968f2bba98eaf1e19909ac78ebf/Template/Variable/DomElementVariable.php#L41
https://github.com/matomo-org/tag-manager/blob/e3c5a5ecd5b1e968f2bba98eaf1e19909ac78ebf/Template/Variable/DomElementVariable.php#L51
https://github.com/matomo-org/tag-manager/blob/e3c5a5ecd5b1e968f2bba98eaf1e19909ac78ebf/Template/Variable/EtrackerConfigurationVariable.php#L128

On the other hand, they are translated in the [`MatomoConfigurationVariable`](
https://github.com/matomo-org/tag-manager/blob/e3c5a5ecd5b1e968f2bba98eaf1e19909ac78ebf/Template/Variable/MatomoConfigurationVariable.php).

Regarding tags and triggers exceptions seem to be never translated. ([`GoogleAnalyticsUniversalTag.php`](https://github.com/matomo-org/tag-manager/blob/e3c5a5ecd5b1e968f2bba98eaf1e19909ac78ebf/Template/Tag/GoogleAnalyticsUniversalTag.php), [`ShareaholicTag.php`](https://github.com/matomo-org/tag-manager/blob/e3c5a5ecd5b1e968f2bba98eaf1e19909ac78ebf/Template/Tag/ShareaholicTag.php),  [`TagsProvider.php`](https://github.com/matomo-org/tag-manager/blob/e3c5a5ecd5b1e968f2bba98eaf1e19909ac78ebf/Template/Tag/TagsProvider.php), [`ThemeColorTag.php`](https://github.com/matomo-org/tag-manager/blob/e3c5a5ecd5b1e968f2bba98eaf1e19909ac78ebf/Template/Tag/ThemeColorTag.php), [`ZendeskChatTag.php`](https://github.com/matomo-org/tag-manager/blob/e3c5a5ecd5b1e968f2bba98eaf1e19909ac78ebf/Template/Tag/ZendeskChatTag.php), [`ElementVisibilityTrigger.php`](https://github.com/matomo-org/tag-manager/blob/e3c5a5ecd5b1e968f2bba98eaf1e19909ac78ebf/Template/Trigger/ElementVisibilityTrigger.php), [`TriggersProvider.php`](
https://github.com/matomo-org/tag-manager/blob/e3c5a5ecd5b1e968f2bba98eaf1e19909ac78ebf/Template/Trigger/TriggersProvider.php))

Template/Variable/CustomRequestProcessingVariable.php
AltamashShaikh: @EPNW-Eric okay but the issue here is someone can write a function like below and it can be a hard rule to write it in specific way, you can maybe, trim the string first and check the first occurrence of `function` and check `(){}` are present or not
```
function(req)
{
}
```

EPNW-Eric: The code in `CustomJsFunctionVariable.php` (see above) checks only the presence of the `function` keyword and no `(`,`)` nor `{`.

What would the rule be to implement here? What about this:
1. the keyword `function`
2. an arbitrary amount of whitespaces, carriage returns, linebreaks and tabs(` `,`\r`,`\n`,`\t`)
3. a `(`
4. an arbitrary amount of whitespaces, carriage returns, linebreaks and tabs
5. the word `request`
6. an arbitrary amount of whitespaces, carriage returns, linebreaks and tabs
7. a `)`

If you have any other suggestions/comments please let me know so I can implement it 😃.

AltamashShaikh: @EPNW-Eric We can add something like this https://3v4l.org/HXnVM
This check should be sufficient to alert the user if they are not using the correct syntax

EPNW-Eric: The  validation at https://3v4l.org/HXnVM does not check if the function is callable with an argument. But that's the difference to a `CustomJsFunctionVariable`: A `CustomRequestProcessingVariable` ***must*** have one argument, otherwise, in the browser there will be a JavaScript exception. If it wasn't for that argument we could just use the validation logic from `CustomJsFunctionVariable`.

I think validating that an argument is there is actually quite hard, so thats why in my initial approach I forced the name of the argument to `request` and furthermore also forced the whole method signature to be `function(request){`.

With respect to the coding style considarations you made and what is already there in `CustomJsFunctionVariable` I think a good compromise would be:
```php
if (strpos($value, 'function(request)') !== 0) {
```
It guarantees that `$value` takes an argument, what is the important part here, even if we have force its name to `request`, and also allows what you mentioned earlier:
```js
function(request)
{
}
```
and additionally works on onliners:
```js
function(request){}
```

AltamashShaikh: @EPNW-Eric I think we are thinking too much on this, I had another look and what you have should be okay, as we are mentioning the same in description
Screenshot from 2023-09-13 18-22-56

AltamashShaikh:
@EPNW-Eric are you done with all your changes? I can have a look tomorrow again if its done

EPNW-Eric:
> @EPNW-Eric are you done with all your changes? I can have a look tomorrow again if its done

Yes, I think it's in a state I'm happy with 😄. Looking forward to hear from you 😃.

AltamashShaikh (approved):
Looks good, I am not sure with the Migration file name, which we can fix it after having a word  internally with the core team

Template/Variable/CustomRequestProcessingVariable.php
snake14: We'll probably want to translate those at some point, but it's fine as they are for now :+1:

snake14 (approved):
Thank you for all the good work @EPNW-Eric . The code looks good and the little bit of functional testing that I did worked as expected :+1:

AltamashShaikh:
@EPNW-Eric We have merged your changes and its should be part of Matomo 5 :+1: 
Thank your for creating this PR :smile:

PR #620, opened 2023-04-14
Reverting Tag Manager JavaScript Changes

snake14:
### Description:

Reverting Tag Manager JavaScript Changes because they appear to be preventing some tracking requests.

### Review

* [ ] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
* [ ] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [ ] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [ ] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)
* [ ] [Code review done](https://developer.matomo.org/guides/pull-request-reviews#code-review-done)
* [ ] [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)
* [ ] [Reviewed for breaking changes](https://developer.matomo.org/guides/pull-request-reviews#reviewed-for-breaking-changes)
* [ ] [Developer changelog updated if needed](https://developer.matomo.org/guides/pull-request-reviews#developer-changelog-updated-if-needed)
* [ ] [Documentation added if needed](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)
* [ ] Existing documentation updated if needed

PR #723, opened 2023-11-17
Improvements to make MTM and _paq work better together

snake14:
### Description:

We're trying to make the MTM tracking code changes from a few months ago work. This re-applies them, making some improvements and adjusting for recent changes. Re-applying PRs 594 and 599.

### Review


Template/Tag/MatomoTag.web.js
```diff
@@ -118,43 +143,230 @@
...
-                    if (matomoConfig.disableCookies) {
+                    if (matomoConfig.disableCookies || disableCookiesIndex !== -1) {
                         tracker.disableCookies();
+                        markIndexAsConfig(disableCookiesIndex);
                     }
 
-                    if (matomoConfig.enableCrossDomainLinking) {
+                    if (matomoConfig.enableCrossDomainLinking || enableCrossDomainLinkingIndex !== -1) {
                         tracker.enableCrossDomainLinking();
-                        tracker.setCrossDomainLinkingTimeout(matomoConfig.crossDomainLinkingTimeout);
```
AltamashShaikh, line 139: @snake14 Why did we removed this line ?

snake14: Good catch. I must have accidentally left it out while reworking all of the changes. I just pushed a change adding it back in.

AltamashShaikh (commented):
@snake14 I have tested this changes and looks good :tada: 

I just noticed 1 thing is that the order of goals was not consistent
Screenshot from 2023-12-06 08-25-43

snake14:
> @snake14 I have tested this changes and looks good 🎉
> 
> I just noticed 1 thing is that the order of goals was not consistent 

Thank you @AltamashShaikh I think that we're probably alright. This was the case even when I tested with the plain JS tracker by itself. I loaded the page multiple times and the goal conversion order changed each time.

AltamashShaikh (approved):

AltamashShaikh:
@tsteur Can you please do an additional review as it created data loss issues last time when we implemented this

snake14:
@tsteur Would you be able to review this? If not, we can merge it after the holidays and have QA test it.

tsteur:
@AltamashShaikh @snake14 Is there a way to see the diff in comparison to the previous solution? And do we have somewhere the information which use case it broke last time? It'll be hard for me to test all the different possible scenarios as it would take quite a long time. I wonder if it otherwise makes more sense to get help from support or so.

snake14:
@tsteur I don't think that there's a clean diff, but here's [the PR that reverted the original changes](https://github.com/matomo-org/tag-manager/pull/620). A diff wouldn't be terribly helpful as there have been a fair number of new configs added to plugins/TagManager/Template/Tag/MatomoTag.web.js since the original ticket. The main issue that I found and fixed was only visible when there were a good number of `_paq.push` calls for configs mixed in with other tracking requests so that at least one tracking request was dropped. It stemmed from me trying to remove the configs from the array and then processing the remaining requests. I changed it to leave the configs in the array, but simply keep track of their indexes and so that we could skip them while processing the tracking requests. I created a few example HTML files as part of this PR to help show some ways to test and confirm the issue has been resolved.
We were planning on doing QA/support testing after your review, but we can move forward to that if you prefer.

tsteur:
@snake14 it be great if others tested first and if needed I could check last 👍

snake14:
@achakko Do you have an environment where you can test this without me merging it into `5.x-dev`?

snake14:
@matomo-org/core-team I received the OK from Product to merge this PR. I merged in the most recent changes from 5.x-dev and everything looks good. Can we please tag this change for the Matomo 5.1.0 release?

PR #747, opened 2024-01-10
Added code to sort by descending order during publish, #PG-3142

AltamashShaikh:
### Description:

Added code to sort by descending order during publish
Fixes: #PG-3142

### Review


snake14 (commented):
The code changes look good. During functional testing Tags, Triggers, and Variables are now sorted by descending modified date. My only concern is that it's grouping by type (Tags, Triggers, and Variables). So, you could have a bunch of tags that were updated months ago at the top and a Trigger or Variable that you just updated at the bottom of the list. I wonder if we should combine the types together so that everything is truly sorted by modified date?

AltamashShaikh:
> The code changes look good. During functional testing Tags, Triggers, and Variables are now sorted by descending modified date. My only concern is that it's grouping by type (Tags, Triggers, and Variables). So, you could have a bunch of tags that were updated months ago at the top and a Trigger or Variable that you just updated at the bottom of the list. I wonder if we should combine the types together so that everything is truly sorted by modified date?

@snake14 Moved to client side sorting to solve the issue completely and removed all the extra code too

vue/src/Version/VersionEdit.vue
```diff
@@ -85,31 +85,31 @@
             <h3>{{ translate('TagManager_ChangesSinceLastVersion') }}:</h3>
             <table v-content-table>
               <thead>
-                <tr>
```
snake14: It's definitely not a deal breaker, but shouldn't these lines be indented within the parent element? Same with the lines below?

snake14 (approved):
Nice work! The sort performs as expected. I just left an optional code style comment.

vue/src/Version/VersionEdit.vue
AltamashShaikh: Removed the unwanted formatting

PR #756, opened 2024-02-16
Automatically delete containers for non existing sites

AltamashShaikh:
### Description:

Added code to delete containers for non existing site
Fixes: #PG-2573

To test you can run this command `./console scheduled-tasks:run --force "Piwik\Plugins\TagManager\Tasks.deleteContainersForNonExistingSite"`

### Review

### DIFF
--- a/Tasks.php
+++ b/Tasks.php
@@ -8,6 +8,9 @@
 namespace Piwik\Plugins\TagManager;
 
 use Piwik\Plugin;
+use Piwik\Plugins\TagManager\Dao\ContainersDao;
+use Piwik\Site;
+use Piwik\Exception\UnexpectedWebsiteFoundException;
 
 class Tasks extends \Piwik\Plugin\Tasks
 {
@@ -24,6 +27,7 @@ class Tasks extends \Piwik\Plugin\Tasks
     public function schedule()
     {
         $this->hourly('regenerateReleasedContainers');
+        $this->daily('deleteContainersForNonExistingSite');
     }
 
     public function regenerateReleasedContainers()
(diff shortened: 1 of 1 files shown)


Tasks.php
```diff
@@ -34,4 +38,42 @@ public function regenerateReleasedContainers()
...
+            return;
+        }
+        $containerDao = new ContainersDao();
+        $containers = $containerDao->getAllContainers();
+        $siteIdsDeleted = [];
+        foreach ($containers as $container) {
+            $idsite = $container['idsite'];
+            if ($idsite === 286) {
+                $a=1;
+            }
```
snake14: Is this left over from testing? It doesn't appear to do anything.

```suggestion
```

Tasks.php
```diff
@@ -34,4 +38,42 @@ public function regenerateReleasedContainers()
...
+            if (!isset($siteIdsDeleted[$idsite]) && !$this->siteExists($idsite)) {
+                $siteIdsDeleted[$idsite] = $idsite;
+            }
+        }
+
+        if (!empty($siteIdsDeleted)) {
+            $tagManager = new TagManager();
+            foreach ($siteIdsDeleted as $siteId) {
+                $tagManager->onSiteDeleted($siteId);
```
snake14, line 62: It looks like this is eventually only [doing a soft delete](https://github.com/matomo-org/tag-manager/blob/be46f9bdcccda167d5922bb04765ae070ca9fe77/Dao/ContainersDao.php#L181). This means that it's going to try deleting the same containers each time because the [getAllContainers](https://github.com/matomo-org/tag-manager/blob/f676ae1076d7601d587335966767ed46f56a62f2/Tasks.php#L50) method returns soft deleted containers. Maybe use the `getActiveContainersInfo` method instead since it only gets active containers and only selects two columns?

snake14 (commented):
Looks good overall and functional testing went well. I just had a couple comments.

Tasks.php
AltamashShaikh: @snake14 Sorry I was testing and pushed this

Tasks.php
AltamashShaikh: @snake14 Good point, updated to use `getActiveContainersInfo` :+1:

snake14 (approved):
Thank you for making those changes. As long as we are sure we want to do a soft delete and not hard delete, everything is working great :+1:

AltamashShaikh:
> Thank you for making those changes. As long as we are sure we want to do a soft delete and not hard delete, everything is working great 👍

Yes, since we have same behaviour for `onSiteDeleted` event, we should keep it same in this case also

PR #789, opened 2024-05-08
Added code to flush localPaq if no callback set but trackPageView set, #PG-3463

AltamashShaikh:
### Description:

Added code to flush localPaq if no callback set but trackPageView set
Fixes: #PG-3463

### Review

### DIFF
--- a/Template/Tag/MatomoTag.web.js
+++ b/Template/Tag/MatomoTag.web.js
@@ -22,15 +22,13 @@
     var localPaq = window._paq && window._paq.length ? JSON.parse(JSON.stringify(window._paq)) : localPaq || [];
     var indexesOfConfigs = [];
     var hasTrackPageViewViaPaq = false;
-    var j = -1;
     // Clear window._paq to prevent things from being tracked too early
     while (window._paq.length > 0) {
-        j++;
+        var poppedItem = window._paq.pop();
         // Check if trackPageView is set in _paq variable to ensure tracking works even with no callback
-        if (Array.isArray(window._paq[j]) && window._paq[j].toString() == ['trackPageView'].toString()) {
+        if (Array.isArray(poppedItem) && poppedItem.toString() == ['trackPageView'].toString()) {
             hasTrackPageViewViaPaqIndex = true;
         }
-        window._paq.pop();
     }
 
     var hasProcessedRemainingTrackings = false;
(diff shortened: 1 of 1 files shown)


Template/Tag/MatomoTag.web.js
```diff
@@ -21,8 +21,15 @@
...
+    var hasTrackPageViewViaPaq = false;
+    var j = -1;
     // Clear window._paq to prevent things from being tracked too early
     while (window._paq.length > 0) {
+        j++;
+        // Check if trackPageView is set in _paq variable to ensure tracking works even with no callback
+        if (Array.isArray(window._paq[j]) && window._paq[j].toString() == ['trackPageView'].toString()) {
+            hasTrackPageViewViaPaqIndex = true;
+        }
         window._paq.pop();
```
snake14: I think what you have would be checking elements at the beginning while the elements are being removed from the end, which means that the end elements wouldn't be checked. How about something like the following?
```suggestion
    var hasTrackPageViewViaPaq = false;
    // Clear window._paq to prevent things from being tracked too early
    while (window._paq.length > 0) {
        var poppedItem = window._paq.pop();
        // Check if trackPageView is set in _paq variable to ensure tracking works even with no callback
        if (Array.isArray(poppedItem) && poppedItem.toString() == ['trackPageView'].toString()) {
            hasTrackPageViewViaPaqIndex = true;
        }
```

snake14 (commented):
Looks like it should work for cases where no MTM pageview trigger applies to the current page. I just had one concern that I commented about.

Template/Tag/MatomoTag.web.js
AltamashShaikh: @snake14 makes sense :+1:

snake14: If this change makes sense and you apply it, I approve of this PR :+1:

Template/Tag/MatomoTag.web.js
```diff
@@ -21,9 +21,14 @@
...
     var localPaq = window._paq && window._paq.length ? JSON.parse(JSON.stringify(window._paq)) : localPaq || [];
     var indexesOfConfigs = [];
+    var hasTrackPageViewViaPaq = false;
     // Clear window._paq to prevent things from being tracked too early
     while (window._paq.length > 0) {
-        window._paq.pop();
+        var poppedItem = window._paq.pop();
+        // Check if trackPageView is set in _paq variable to ensure tracking works even with no callback
+        if (Array.isArray(poppedItem) && poppedItem.toString() == ['trackPageView'].toString()) {
+            hasTrackPageViewViaPaqIndex = true;
```
snake14: I just realised that the variable names weren't consistent. Sorry that I missed that earlier.
```suggestion
            hasTrackPageViewViaPaq = true;
```

AltamashShaikh: @snake14 Updated please check

snake14:
Closing as we're rolling back the changes with #790

PR #871, opened 2024-09-02
Increase the allowed length of name fields

### LINKED ISSUE #816: Increase tags and triggers title length
Hi,

Actually, the tags and triggers title field is limited to only 50 characters.
Can you increase this limit to 255 characters ? It would be a great improvement and allow users to follow naming guideline such as :

Matomo - Event - my_custom_event_name_even_if_the_name_is_long

We actually can't name a tag with a long name, and it'll force users to change the naming convention or change event name to shorts (which can cause confusion).

Regards,
Ronan HELLO

snake14:
### Description:

It has been requested that we increase the allowed length of tag and trigger names, so that's what this PR does. We are increasing the maximum name length from 50 to 255 for tags, triggers, variables, and containers. The version max length is being increased in the database, but we're leaving the 30 max restriction in the UI. This is partially because there isn't really a need for longer version names, plus the version name is included in the container JS file and having a long one increases the size of that file.

Fixes: #816 

### Review


Menu.php
```diff
@@ -91,7 +91,8 @@ public function configureTagManagerMenu(MenuTagManager $menu)
             if (!empty($container)) {
 
                 $params = array('idContainer' => $idContainer); // not needed as it is already present in url but we make sure the id is set
-                $menuCategory = $container['name'];
+                $menuCategory = strlen($container['name']) > 50 ? substr($container['name'], 0, 50) . '...' : $container['name'];
```
AltamashShaikh: Better to use an ellipsis here instead of 3 dots
```suggestion
                $menuCategory = strlen($container['name']) > 50 ? substr($container['name'], 0, 50) . '…' : $container['name'];
```

vue/src/Tag/TagList.vue
```diff
@@ -258,6 +263,7 @@ export default defineComponent({
   data(): TagListState {
     return {
       hasWriteAccess: Matomo.hasUserCapability('tagmanager_write'),
+      triggerTruncateLength: 40,
```
AltamashShaikh, line 266: @snake14 This should be 50 ?

AltamashShaikh (commented):
@snake14 Overall looks good, check my 1 comments around triggerTruncateLength, rest looks good to me.

Menu.php
snake14: Thank you :+1:

vue/src/Tag/TagList.vue
snake14: @AltamashShaikh I tried 50 at first, but I thought that 40 looked better. There was some strange wrapping behaviour when I tried 50.

AltamashShaikh (approved):
Looks good :+1: 
Just resolve the merge conflicts and merge the PR

PR #879, opened 2024-09-11
Added the option to sync from the GTM dataLayer

### LINKED ISSUE #304: Google Tag Manager type of dataLayer doesn't triggers in Matomo Tag Manager
Hello,

I have a problem with the detection of Google Tag Manager dataLayer in Matomo. 
The detection work when the dataLayer is sent when a page is loaded but when the dataLayer is sent on an interaction like a click on a button the dataLayer is not detected by Matomo Tag Manager.
The dataLayers are based on the Google Tag Manager type
For example
window.dataLayer = window.dataLayer || [];
window.dataLayer.push({
'event': 'test'
});

Is it a bug or the way it works with GTM dataLayers in Matomo Tag Manager ?
Do you have a solution for Matomo Tag Manager to detect GTM dataLayers when they are sent after the page load ?

We are operating a migration from GA to Matomo and we want to avoid changing the dataLayers on the website.

Thanks in advance.

snake14:
### Description:

Added the option to sync from the GTM dataLayer. The new setting will default to off for all existing containers and on for any new containers. This is so that the behaviour won't suddenly change for existing customers.

Fixes: #304 

### Review


javascripts/tagmanager.js
```diff
@@ -1830,6 +1830,20 @@
                 }
             }
 
+            // Only sync the dataLayer changes from GTM if the config has been set
+            if (!('undefined' === typeof activelySyncGtmDataLayer) && activelySyncGtmDataLayer && 'undefined' !== typeof windowAlias.dataLayer && utils.isArray(windowAlias.dataLayer)) {
```
AltamashShaikh: @snake14 this check needs to be updated, otherwise it won't work if someone tries to initialise dataLayer after the script is loaded

Case when adding dataLayer via console
Screenshot from 2024-09-12 11-46-48

With suggested change
Screenshot from 2024-09-12 11-48-25


```suggestion
            if (!('undefined' === typeof activelySyncGtmDataLayer) && activelySyncGtmDataLayer) {
                windowAlias.dataLayer = windowAlias.dataLayer || [];
```

AltamashShaikh (changes requested):
Overall looks good, left 1 comment

javascripts/tagmanager.js
snake14: Thank you for the good suggestion. I have applied the change.

AltamashShaikh (approved):
Looks good :+1:

PR #887, opened 2024-09-20
Created new setting to restrict access to MTM

### LINKED ISSUE #602: Hide Tag Manager Section
We have had a customer ask about hiding the Tag Manager section for users that have '**View**' access.
Currently users with **View** access can access the Tag Manager and see the Tag/triggers that are in place. (they cannot proceed to edit or see the contents of the tags/triggers.) 

However it has been requested if it's possible to implement an option that could be set to 'hide' the Tag Manager section completely (if the super user has chosen too) if a user has **View** access.

Comment from customer: ": how can we give View Access to a Matomo dashboard to somebody without giving them view access of Tag Manager as well? As of right now, whenever we sent view access to our clients, they can also see the Tag Manager section, which we would want to avoid."

snake14:
### Description:

Created new setting to restrict access to MTM. It's a new dropdown in the MTM section of General Settings. It allows the selection of the minimum level of permission with access to MTM. When a user doesn't have access, the top navigation is removed and all controller and API methods return an error message.

Fixes: #602 

### Review


AltamashShaikh (changes requested):
@snake14 Good work, but I found this feature breaking the Measurables > Settings page if we set the below setting. I assume similar would be the issue for No Data Screen

Screenshot from 2024-09-23 07-29-51

Screenshot from 2024-09-23 07-28-01

Screenshot from 2024-09-23 07-28-07

snake14:
> @snake14 Good work, but I found this feature breaking the Measurables > Settings page if we set the below setting. I assume similar would be the issue for No Data Screen

Good catch @AltamashShaikh . I wondered why there was no permission checks on a couple of the API methods. That's probably why. I could remove the the new validation from those methods, but I wonder if we should hide those MTM sections that make the API call in the first place. I wouldn't think that a user that doesn't have MTM access should be installing the tracking code. What do you think @Stan-vw ?

snake14:
> > @snake14 Good work, but I found this feature breaking the Measurables > Settings page if we set the below setting. I assume similar would be the issue for No Data Screen
> 
> Good catch @AltamashShaikh . I wondered why there was no permission checks on a couple of the API methods. That's probably why. I could remove the the new validation from those methods, but I wonder if we should hide those MTM sections that make the API call in the first place. I wouldn't think that a user that doesn't have MTM access should be installing the tracking code. What do you think @Stan-vw ?

@AltamashShaikh I went ahead and hid the MTM code in those places.

SystemSettings.php
```diff
@@ -21,6 +23,16 @@ class SystemSettings extends \Piwik\Settings\Plugin\SystemSettings
     const CUSTOM_TEMPLATES_ADMIN = 'admin';
     const CUSTOM_TEMPLATES_SUPERUSER = 'superuser';
 
+    const USER_PERMISSON_LIST = [
+        \Piwik\Access\Role\View::ID,
```
AltamashShaikh, line 27: Can we use name index ?
```suggestion
       'view' => \Piwik\Access\Role\View::ID,
```

SystemSettings.php
```diff
@@ -31,10 +43,68 @@ class SystemSettings extends \Piwik\Settings\Plugin\SystemSettings
...
     }
 
+    private function createRestrictAccessSetting(): SystemSetting
+    {
+        return $this->makeSetting('restrictTagManagerAccess', \Piwik\Access\Role\View::ID, FieldConfig::TYPE_STRING, function (FieldConfig $field) {
+            $field->title = Piwik::translate('TagManager_SettingRestrictAccessTitle');
+            $field->uiControl = FieldConfig::UI_CONTROL_SINGLE_SELECT;
+            $field->description = Piwik::translate('TagManager_SettingRestrictAccessDescription');
+            $field->availableValues = [
+                self::USER_PERMISSON_LIST[0] => Piwik::translate('TagManager_SettingRestrictAccessView'),
```
AltamashShaikh: Similarly
```suggestion
                self::USER_PERMISSON_LIST['view'] => Piwik::translate('TagManager_SettingRestrictAccessView'),
```

SystemSettings.php
```diff
@@ -31,10 +43,68 @@ class SystemSettings extends \Piwik\Settings\Plugin\SystemSettings
...
+            return true;
+        }
+
+        $settingValue = $this->restrictTagManagerAccess->getValue();
+
+        // We need to allow checks with no site ID since we might be in the Administration section
+        if ($idSite === 0) {
+            switch ($settingValue) {
+                // View
+                case self::USER_PERMISSON_LIST[0]:
```
AltamashShaikh:
```suggestion
                case self::USER_PERMISSON_LIST['view']:
```

AltamashShaikh (commented):
Looks good to me left 1 suggestion to improve code readability

SystemSettings.php
snake14: @AltamashShaikh I thought about doing that, but wanted to ensure the order of the array since I use the index number for comparison during the security check. If I used a name index, I would probably use a separate array to map the index name to the permission weight; that just felt like complicating things more than necessary.

snake14: It also felt a little funny creating an array like ['view' => 'view', 'write' => 'write', ...];

AltamashShaikh: What if we make individual constants ?

AltamashShaikh (approved):

SystemSettings.php
snake14: I could create constants, but I don't see much point as the constants already exist in other classes. I just thought that the array was a good way to track permission weight.

I suppose that I could use `array_search()` to return the index of the array instead of hard coding the int indexes. That might help with maintainability.

snake14: I created a helper function that looks up the index of the string. I think that makes the code a bit easier to read at a glance :+1:

PR #924, opened 2024-11-14
Implement option to copy a tag

### LINKED ISSUE #615: [Enhancement] Add "Copy tag" option in Matomo Tag Manager
Sometimes we need to build a really similar Tag / Trigger within Matomo TMS. When it happens, being able to just copy a tag would do a fine addition to Matomo Tag Manager features !

## Summary

Considering this feature, it would require to add a "copy" button on the side of each Tag as you can see in the following exemple : 

image

Which would lead to the following screen : 

image

It would create the same tag, named "Copy of MTM - myTagName" with either : 
 - the same triggers if "No" was selected on the previous screen 
 - the copy of the triggers if "Yes" was selected on the previous screen

Of course this goes with the possibility to copy triggers within the triggers section directly as well.

## Your Environment

* Matomo Version: Matomo Cloud

snake14:
### Description:

This PR implements the ability to copy tags. It tries to reuse existing triggers and variables when copying references from the tag. If no matches exist, any dependencies will be copied as well.

Fixes: #615 

### Review


Dao/ContainerReleaseDao.php
```diff
@@ -168,6 +168,11 @@ public function deleteNoLongerExistingEnvironmentReleases($availableEnvironments
         return $query->rowCount();
     }
 
+    protected function isNameAlreadyUsed(int $idSite, string $name, ?int $idContainerVersion = null): bool
+    {
+        return true;
```
AltamashShaikh, line 174: @snake14 Did we forgot to add some logic to check the name here ?

vue/src/Tag/TagList.vue
```diff
@@ -135,7 +135,7 @@
               <span>{{ tag.updated_date_pretty }}</span>
             </td>
             <td
-              class="action"
+              :class="getActionClasses"
               v-show="hasWriteAccess"
```
AltamashShaikh, line 139: @snake14 Should we check `hasWriteAccess && hasWriteCapability` although it should be covered by `getActionClasses` check I beleieve

AltamashShaikh (changes requested):
@snake14 
1. After copying a tag, there is no notification that its a success or failure.
2. For tags when we click copy the default container should be the current container

I also faced 1 issue of copy tag throwing and error "This name is already in use" when copying a same tag twice in an another container

AltamashShaikh (commented):
@snake14 The copy notification is getting hidden, if preview mode is enabled

Without preview mode

image

With preview mode, no notification message is displayed

image

vue/src/Tag/TagList.vue
snake14: @AltamashShaikh Correct. Adding a check here would be redundant. It's already handled in `getActionClasses` where it uses the `hasPublishCapability` function to check for write and custom templates.

Dao/ContainerReleaseDao.php
snake14: @AltamashShaikh No. Releases don't have a name, so this dao doesn't use the method, but has to implement it since it's abstract. I added a comment to explain why the return value is hard coded.

snake14:
Thank you for the good review @AltamashShaikh .

It looks like I'm going to have to refactor things so that the notifications are handled via Vue instead of PHP, because the PHP notification just doesn't appear to be reliable enough.

I thought that I had it selecting the current container by default, but I'll make sure.

I have test cases around copying to the same location twice, but I guess I only did that for triggers and variables. I'll add one for tags too and fix the name issue. :+1:

AltamashShaikh (approved):
Looks good now :+1:

Issue #990, opened 2025-02-25
Publish container not working for user with write access & getting no success notification when creating a new version/ publishing a container


tsteur:
* have a user with write and publish live permission

![Image](https://github.com/user-attachments/assets/10554f48-6e90-45d1-9b85-227a984dc645)

* With this user, edit a container that has custom templates
* Edit a tag (that doesn't have custom templates)
* Create a new version and publish release
* The version is created but there's no feedback that a new version has been created/published.
* I try to click again...
* Then I get an error that "The name is already in use" meaning the publish worked before.
* I open the screen with "Changes since last version" and can see that no changes actually have been published

![Image](https://github.com/user-attachments/assets/dad3da9d-24fc-4c8c-a233-a606b2c45ebc)

* Neither do the published versions if an environment assigned (likely explains why the changes since last version always shows up)

![Image](https://github.com/user-attachments/assets/99768f4e-9289-4239-b306-77de469c159d)

This works as a super user. But for a user with write access.
Expected: Get a notification on success, close the modal, assign the correct environment, actually publish the container, (optionally reload the list of published versions to show the newly published version if that page was open previously

snake14:
Thank you for taking the time to create this issue @tsteur . Can you please confirm that the changes were in fact not published. It appears that the changes were actually published, but the Matomo Configuration variables are showing as different (note the date is almost 2 years ago when there have been many releases since). I'm guessing that a migration didn't run correctly and it's showing as different because there's a new field which wasn't added to the previous version.

When I tested in my local instance, I didn't see the success notification, but the version was published as expected and the changes list was empty afterward.

We definitely need to look into why the success notification doesn't display. We might also want to look into an improvement related to the comparison issue as this keeps popping up when migrations are skipped or don't run correctly.

tsteur:
@snake14 the publishing worked nicely with super user access. It was publishing with the write user but to no environment. So when I tried to publish to live again a 2nd time, it would show the same changes again.
By the looks the updates are executed. To be safe, I re-ran all updates and tested again and same issue

Published, then version without environment was released:
![Image](https://github.com/user-attachments/assets/8133dd04-1016-4e4f-8bcd-5b77fb3d016a)

Then I click on publish again a 2nd time and same changes show up again

![Image](https://github.com/user-attachments/assets/d4eae790-051b-4597-a8f8-6b481d2961bb)

It seems the version entry is created, but not the release entry. 

Looking at the request information there is an error:

[...]

PR #1007, opened 2025-04-24
Handles case where user has access to site, but not the idsite mapped under Matomo Configuration Variable, #PG-4170

AltamashShaikh:
## Description
The container publish action was throwing an exception as the user had access to the site, but no access on the site mapped under Matomo Configuration Variable

## Issue No
#PG-4170, #990

1. Login as super user.
2. Create a new container.
3. Add a new Matomo tag under this container.
4. Create a new Matomo Configuration variable and add a different siteId from the site you are logged in.
5. Create an admin user and give access to the site where we created a new container.
6. Ensure admin had publish capability.
7. Now login as admin and try publishing the container, this will throw an error in background.
8. Checkout this PR and this should start working.

## Checklist
- [NA] Version bumped?


tests/Integration/APITest.php
```diff
@@ -439,6 +441,25 @@ public function test_deleteContainerVersion_shouldFailWhenHavingOnlyWritePermiss
...
+        $variable = $variables[0];
+        $parameters = $variable['parameters'];
+        $parameters['idSite'] = $newIdSite;
+        $this->api->updateContainerVariable($idSite, $idContainer, $idContainerDraftVersion, $variable['idvariable'], $variable['name'], $parameters);
+        $this->setAdminUser();
+        FakeAccess::$idSitesCapabilities = array(UseCustomTemplates::ID => array($idSite));
+        Site::clearCache(); // so that cache has only the sites admin has access, else was unable to add a test for #990
+        $version = $this->api->createContainerVersion($idSite, $idContainer, 'foo', '', $idContainerDraftVersion);
+        $this->assertNotEmpty($version);
+    }
```
snake14, line 461: I'm not sure if this test case completely covers the changes in this PR. In my testing, it appeared that the version was created even before this PR, but it failed to publish the new version.

snake14 (commented):
Looks really good. I confirmed that publishing a Matomo Config variable which tracks to a site my user didn't have access to failed before this PR and succeeded after :+1: 

I only had a comment about the new test case.

tests/Integration/APITest.php
AltamashShaikh: @snake14 I tried doing this
- Line 445-455 login as a super user and try creating a new Container and change the idsite to a different idSite for that container.
- Line 456-460, try releasing a container, without access to the new site mapped in Matomo Configuration Variable
- This throws an exception and is handled via the new code we added in catch block.

image

snake14: Thank you for the clarification. The new test case is more granular than I realised and covers the root cause of the failed publish :+1:

snake14 (approved):

Issue #929, opened 2024-11-22
Missing Explanation for History Change Trigger in Matomo Tag Manager


Chardonneaur:
Hi team,

There is a missing ? for History Change trigger within Matomo Tag Manager. As a result, people don't know what this trigger is about.

![Screenshot from 2024-11-22 10-44-15](https://github.com/user-attachments/assets/3cb3d2f3-d46b-46c8-91c4-6d8c4664fad5)

---

### Steps to reproduce

1. Open **Matomo Tag Manager** and navigate to the **Triggers** section.
2. Attempt to create a new trigger using the **History Change** trigger type.
3. Review the options and settings for this trigger.

### What I got:

The **History Change** trigger lacks a clear explanation or help text (e.g., a question mark with a tooltip) to describe its purpose and how it works. This creates confusion for users trying to understand when and how to use it.

### What I expected instead:

A **help icon** (e.g., a question mark or tooltip) next to the **History Change** trigger that provides a brief explanation of its function, such as:

> *"The History Change trigger activates when the browser's history changes, such as when using single-page applications (SPAs) where the URL changes without a full-page reload."*

---

AltamashShaikh:
@Chardonneaur Thanks for creating this issue, it seems we don't have `HistoryChangeTriggerHelp` key set and that's why there is no `?` mark present.

PR #1010, opened 2025-05-08
Adds missing helptext for triggers, #PG-4006

AltamashShaikh:
## Description
It was noticed that some triggers do not show the `?` icon, which shows the helptext on hover. The PR aims to solve this issue.

## Issue No
#PG-4006, #929

1. Navigate to TagManager
2. Select a Container
3. Click Triggers from left menu
4. You would notice that Pageview and few others will not have a `?` icon.
5. Checkout this branch and you should now see the icon

## Checklist
- [NA] Version bumped?


snake14 (approved):
Nice work! I was able to see that all trigger types have help text now :+1:

Issue #989, opened 2025-02-25
Unclear error message when trying to edit a custom template without permission


tsteur:
* Given you have a user that has write and publish permission tag manager but not for custom templates.
* Try to edit a custom tag
* Click Save
* See error message that shows up:

![Image](https://github.com/user-attachments/assets/b835ff6f-8007-4c4d-bcbf-7514ae3be774)

The error message in red isn't helpful in that situation. Yes there is a yellow warning from the beginning when opening the tag, but this may be missed easily.

The request that fails does have the actual error message that is a bit more readable:

![Image](https://github.com/user-attachments/assets/022709ca-e741-408d-8cc5-a699d7214092)

Expected behavior: The error message from the request should be shown in red.
Alternatively: Don't allow editing of any of the fields and neither allow submitting the form.

snake14:
Thank you for creating this issue @tsteur . I like the idea of making sure that the form can only be edited/submitted when permissions allow.

PR #1019, opened 2025-05-23
Removes save button from custom templates if no access, #PG-4171

AltamashShaikh:
## Description
It was noticed that the save button was visible for user having write access but not custom_template capability, as the user had no custom_template capability, system was throwing an error on save, this PR aims to hide the button from UI, if the user doesn't have access to it.

## Issue No
#PG-4171, #989

1. Login to Matomo as a super user
2. Navigate to TagManager
3. Select a container
4. Create a Custom HTML tag and a Custom Javascript variable
5. Now login as a write_user with no custom_template capability
6. Try to edit the tag and variable created in step4 and click update
7. You will see a red box error
8. Now checkout this PR and you should see the update button is not displayed anymore and the alert box is updated to danger.

## Checklist
- [NA] Version bumped?


snake14 (approved):
Looks good. I confirmed that custom tags and variables show the new hover text and red banners for users which don't have the correct capability. :+1: 

I noticed that the trigger template was updated but couldn't find any trigger types which were affected. I don't think we have any custom template triggers, do we?

AltamashShaikh:
> Looks good. I confirmed that custom tags and variables show the new hover text and red banners for users which don't have the correct capability. 👍
> 
> I noticed that the trigger template was updated but couldn't find any trigger types which were affected. I don't think we have any custom template triggers, do we?

@snake14 Nope, I think its more of copy paste or added so that in future if we add any customTemplate in triggers same logic stays there too, so I added the similar changes for triggers too.

PR #1030, opened 2025-07-22
Update label selectors to not rely on a specific field name or id

michalkleiner:
## Description
A small tweak to target advanced options dropdowns differently not to rely on their explicit names. This relates to core PR https://github.com/matomo-org/matomo/pull/23385 which would cause the IDs to be different and the styling would stop working.

## Issue No
n/a

Change the name of the dropdowns in tracking code advanced options (when choosing tracking code for a measurable using tag manager) and observe the labels get misaligned.

## Checklist
- [✔] Tested locally
- [✔] visual tests still passing
- [NA] Version bumped?

### DIFF
--- a/vue/src/TagmanagerTrackingCode/TagManagerTrackingCode.less
+++ b/vue/src/TagmanagerTrackingCode/TagManagerTrackingCode.less
@@ -14,16 +14,16 @@
     margin: 8px 0 0 0;
   }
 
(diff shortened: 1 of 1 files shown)


AltamashShaikh (approved):
I tested these changes on no-data screen and tracking code generator page in Admin and seems to work as expected. :+1:

PR #972, opened 2025-01-09
Updating reference check to include variables referencing variables

snake14:
### Description:

It was found that the existing references check during the deletion process doesn't include checking if the variable is referenced by CustomJS variables. The PR fixes that.

Fixes: #895 

### Review

* [ ] [Functional review done](https://developer.matomo.org/guides/pull-request-reviews#functional-review-done)
* [ ] [Potential edge cases thought about](https://developer.matomo.org/guides/pull-request-reviews#potential-edge-cases-thought-about) (behavior of the code with strange input, with strange internal state or possible interactions with other Matomo subsystems)
* [ ] [Usability review done](https://developer.matomo.org/guides/pull-request-reviews#usability-review-done) (is anything maybe unclear or think about anything that would cause people to reach out to support)
* [ ] [Security review done](https://developer.matomo.org/guides/security-in-piwik#checklist)
* [ ] [Wording review done](https://developer.matomo.org/guides/pull-request-reviews#translations-wording-review-done)
* [ ] [Code review done](https://developer.matomo.org/guides/pull-request-reviews#code-review-done)
* [ ] [Tests were added if useful/possible](https://developer.matomo.org/guides/pull-request-reviews#tests-were-added-if-usefulpossible)
* [ ] [Reviewed for breaking changes](https://developer.matomo.org/guides/pull-request-reviews#reviewed-for-breaking-changes)
* [ ] [Developer changelog updated if needed](https://developer.matomo.org/guides/pull-request-reviews#developer-changelog-updated-if-needed)
* [ ] [Documentation added if needed](https://developer.matomo.org/guides/pull-request-reviews#documentation-added-if-needed)
* [ ] Existing documentation updated if needed

PR #1035, opened 2025-09-02
Fixes reference check not allowing import of same draft version, #PG-4532

### LINKED ISSUE #1034: Tags/Triggers removed after version restore
**Context:**
When restoring a container version to the same existing container (overwriting an existing container), the import fails to complete with an error "_This variable is in use elsewhere and cannot be deleted. Please remove all references to it and try again._"

Image

Viewing the Dashboard for the container after this error is displayed, all tags and triggers have been deleted from the container.

Image

**Steps to Reproduce:**  

This can be replicated by restoring a container version over an already existing container.
When managing a container select _Versions > Export Version_.

Image

Select Import on the same screen and paste in the contents of the exports .json file for your container.

Image

Select _Overwrite current draft with this version._

The error _"This variable is in use elsewhere and cannot be deleted. Please remove all references to it and try again."_ is displayed and exisiting tags/triggers are removed from the container.

This was replicated on a Matomo Cloud subscription.

AltamashShaikh:
## Description
Fixes reference check not allowing import of same draft version
Fixes: #1034 

## Issue No
#PG-4532

1. Go to TagManager
2. Select the container
3. Select versions
4. At bottom of the page click "Export draft"
5. Now import the same draft and it will not work
6. Checkout this PR and it should allow import.

## Checklist
- [NA] Version bumped?

### DIFF
--- a/API/Import.php
+++ b/API/Import.php
@@ -129,11 +129,11 @@ class Import
         }
 
         foreach ($this->triggers->getContainerTriggers($idSite, $idContainerVersion) as $trigger) {
-            $this->triggers->deleteContainerTrigger($idSite, $idContainerVersion, $trigger['idtrigger']);
+            $this->triggers->deleteContainerTrigger($idSite, $idContainerVersion, $trigger['idtrigger'], true);
         }
 
         foreach ($this->variables->getContainerVariables($idSite, $idContainerVersion) as $variable) {
-            $this->variables->deleteContainerVariable($idSite, $idContainerVersion, $variable['idvariable']);
+            $this->variables->deleteContainerVariable($idSite, $idContainerVersion, $variable['idvariable'], true);
         }
 
         $ecv = $exportedContainerVersion;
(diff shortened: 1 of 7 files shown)


AltamashShaikh:
This seems to be a regression from - https://github.com/matomo-org/tag-manager/pull/972

snake14 (dismissed):
I had to import the draft JSON from the Jira ticket and try to replace it with my own draft in order to reproduce the issue. The error went away when I checked out this branch :+1: 

It feels a little risky allowing the check to be skipped, but this seems like the most pragmatic approach as long as we ensure it only gets used when absolutely necessary.

A new test case would be nice, but it might be more effort to set up. It looks like you might be able to copy the [test_import_possible_to_overwrite_itself](https://github.com/matomo-org/tag-manager/blob/4f905213b5c507abfd90f400afc658d0e23ef6be/tests/System/APITest.php#L425) system test case and then make sure that the draft being overwritten has a customJS variable which references another customJS variable with an older created_date (so that the referenced variable is deleted first).

snake14 (approved):
Nice work on the new test case! I was able to check out the 5.x-dev version of API/Import.php to confirm that the test case threw an exception. :+1:

PR #1037, opened 2025-09-04
Added code to trim names and rollback if import results in error, #PG-4400

### LINKED ISSUE #1029: Delete the version if any error occurs before all the relevant tags/triggers/variables are loaded into that version
The process to create a new version is:

- We create a new version from draft
- Export the draft container version.
- Import the exported version into the new version

This flow works fine, but creates an issue when there is some error in dependent variable/tags/triggers.

For example, a customer created a variable name having tab space `GDPR Consent level value	`, this variable was used in triggers, but when the import step was being performed it threw an error `TagManager_ErrorVariableInConditionAtPositionNotFound` and as a result the version had tags as empty array, the customer later manually made this version as live and this broke the tracking on their site.

Ideally the customer, should have tested the draft version via preview/debug mode, but it would be good, if we just delete the recently created version on exception as this half-baked error prone version is shown in the version list and can be publish to live.

AltamashShaikh:
## Description
Added code to trim names and rollback if import results in error

## Issue No
#PG-4400
Fixes: #1029 

1. Go to TagManager
2. Select a container
3. Navigate to versions
4. **Important:** Export draft
5. Now import a version using this JSON
```
{"context":"web", "idcontainer":"aacont1","tags":[{"idtag":2,"type":"CustomHtml","name":"My Tag 2","description":"My Tag 2 description","status":"active","parameters":{"customHtml":"<script><\/script>","htmlPosition":"bodyEnd"},"fire_trigger_ids":[1],"block_trigger_ids":[2],"fire_limit":"once_lifetime","priority":343,"fire_delay":1350,"start_date":"2017-01-02 03:04:05","end_date":"2029-01-02 03:04:05","created_date":"2025-09-04 06:56:46","updated_date":"2025-09-04 06:56:46","created_date_pretty":"Sep 4, 2025 06:56:46","updated_date_pretty":"Sep 4, 2025 06:56:46"}],"triggers":[{"idtrigger":1,"type":"CustomEvent","name":"My trigger1","description":"My trigger1 description setting it to a very long description to show up in the UI clipped","parameters":{"eventName":"foo"},"conditions":[{"comparison":"equals","actual":"Constant-Reference ","expected":false}],"created_date":"2025-09-04 06:56:46","updated_date":"2025-09-04 06:56:46","created_date_pretty":"Sep 4, 2025 06:56:46","updated_date_pretty":"Sep 4, 2025 06:56:46"}],"variables":[{"type":"Constant","name":"Constant-Reference","description":"","default_value":"","lookup_table":[],"parameters":{"constantValue":"test"}},{"type":"CustomJsFunction","name":"Custom JavaScript Constant Ref","description":"","default_value":"","lookup_table":[],"parameters":{"jsFunction":"function () { return \"{{Constant-Reference}}\"; }"}}]}
```
6. You will notice, that it sends an error 
7. Close the modal and view tags, all the tags would be deleted as we delete existing tags/triggers and variables and try to add a variable first, if that results in error we have a wrong draft state
8. Checkout this PR
9. Import the draft exported in step4
10. Perform step 3-7 and now view tags, nothing should be deleted.

Note: On exception we are creating the tags/triggers/variables again so ID would change and we would see all of that as a changed values.

## Checklist
- [NA] Version bumped?


API.php
```diff
@@ -1477,7 +1480,15 @@ public function importContainerVersion($exportedContainerVersion, $idSite, $idCo
 
         $this->containers->checkContainerVersionExists($idSite, $idContainer, $idContainerVersion);
         $this->enableGeneratePreview = false;
-        $this->import->importContainerVersion($exportedContainerVersion, $idSite, $idContainer, $idContainerVersion);
+        try {
+            $this->import->importContainerVersion($exportedContainerVersion, $idSite, $idContainer, $idContainerVersion);
+        } catch (Exception $e) {
+            if (!$isDraftRestoreCall && !empty($draft)) {
+                // rollback to old working draft
```
snake14: Would it be cleaner to delete the backup version created on line 1478? It feels confusing that a new version is created each time even if the import fails. You could simply add `$backupVersionId = ` to line 1478 and do the following:
```suggestion
            if (!$isDraftRestoreCall && !empty($draft)) {
                // Delete the backup since the import failed and is being rolled back
                if (!empty($backupName)) {
                    $this->deleteContainerVersion($idSite, $idContainer, $backupVersionId);
                }
            
                // rollback to old working draft
```

API.php
```diff
@@ -1461,7 +1461,10 @@ public function importContainerVersion($exportedContainerVersion, $idSite, $idCo
             throw new Exception(Piwik::translate('TagManager_ErrorContainerVersionDoesNotExist'));
         }
 
-        $exportedContainerVersion = Common::unsanitizeInputValue($exportedContainerVersion);
+        if (!$isDraftRestoreCall) {
+            $draft = $this->exportContainerVersion($idSite, $idContainer);
+            $exportedContainerVersion = Common::unsanitizeInputValue($exportedContainerVersion);
```
snake14, line 1466: Why is this line necessary when it wasn't before? Is this fixing an additional pre-existing issue?

snake14 (commented):
Nice work! I was able to reproduce the issue and confirm that the PR prevents it. :+1: 

I just had a couple questions.

API.php
AltamashShaikh: @snake14 This line was creating an issue for rollback when doing json_decode, so moved the code to do this only when its not a restore action.

Getting this error
image

API.php
AltamashShaikh: Done :+1:

snake14 (approved):
Thank you for the changes and answering my questions. It works as expected :+1:

API.php
```diff
@@ -1448,7 +1448,7 @@ public function exportContainerVersion($idSite, $idContainer, $idContainerVersio
      * @param string $backupName   If specified, a backup of the current draft will be created under this version name.
      * @return array
      */
-    public function importContainerVersion($exportedContainerVersion, $idSite, $idContainer, $backupName = '')
+    public function importContainerVersion($exportedContainerVersion, $idSite, $idContainer, $backupName = '', $isDraftRestoreCall = false)
```
sgiehl, line 1451: I came across this change due to a UI screenshot being changed in core:

Is this parameter supposed to be public API? If so it would be helpful to clearly describe its purpose at least in the doc block. By just looking at the parameter name I don't have a clue what it's supposed to do.
If it's more like an internal parameter, only used by our UI, that shouldn't be used by API consumers directly, it should be named `$_isDraftRestoreCall`, as that would hide it from the API listing page.

Also it would be awesome if you could start adding type hints for all API parameters where the type can be hinted. At least for new parameters added.

AltamashShaikh: @sgiehl thanks for the info, created - https://github.com/matomo-org/tag-manager/pull/1038

PR #1039, opened 2025-09-18
Adjust MTM active data layer sync to maintain GTM sync

### LINKED ISSUE #1014: The feature to "Actively synchronise from the Google Tag Manager data layer" interferes with GTM to fire events
GTM triggers don't work anymore, if "Actively synchronise from the Google Tag Manager data layer" in Matomo is enabled.

snake14:
## Description
The key issue is that MTM syncing with the GTM data layer prevents changes to the data layer from triggering events in GTM. Thank you to @karlsve for providing a commit of a suggested solution. This is simply taking that commit and minifying the JS so that things can be more easily reviewed and tested.

## Issue No
Fixes: #1014

1. Have GTM and MTM running on the same site
2. Have at least one trigger configured in GTM which triggers when an event is pushed to the data layer.
3. Enable the `Actively synchronise from the Google Tag Manager data layer` setting in the MTM container settings.
4. See that new data layer entries are added to both data layers, but don't trigger corresponding events in GTM.
5. Check out this PR
6. Reload the page and confirm that both MTM and GTM receive any new events pushed to the data layer and that GTM triggers events as expected.

## Checklist
- [✔/✖] Tested locally or on demo2/demo3?
- [NA] Version bumped?


AltamashShaikh (approved):
works for me as expected

PR #1056, opened 2025-11-02
PG-4669 test with matomo5_[max|min]_php [ignore_release]

snake14:
Updating the matomo-tests workflow to use shared min/max PHP versions so that we don't have to create a new PR every time we want to support a new PHP version for Matomo and all plugins.


james-hill-matomo (approved):
Double checked tests

PR #1070, opened 2025-12-02
Do not remove default timeout in UI tests

sgiehl:
## Description

In our code builds we have seen recently that sometime the CI pipeline is hanging when running various tests. This then blocks the job and times out after hours. Most likely this happens as many tests still have `timeout(0)` defined in the beginning, preventing the default timeout to apply and the tests to be abort after a certain time.


james-hill-matomo (approved):
LGTM.

I wonder if we may need to add it back in with a large value (not 0), but given that these tests passed in this PR I'm happy.

Probably better to chop the tests up instead.

PR #1074, opened 2026-01-05
Adds code to check customTemplate capability on destination site if customHTML tag being copied, #AS-455

AltamashShaikh:
## Description
Adds code to check customTemplate capability on destination site if customHTML tag being copied

## Issue No
#AS-455

1. Create a user having Admin access for SiteA and write access for SiteB
2. Now create a CustomHTML tag in siteA
3. Now try to create a customHTML tag in SiteB, it will not allow you due to write access
4. Now try to copy the customHTML tag from SiteA to SiteB and the custom HTML tag will be copied instead of error
5. New checkout this PR and try to copy, it will not succeed

## Checklist
- [NA] Version bumped?


lachiebol (changes requested):
Functionality works and the copy is being stopped. Only one issue, when the copy fails, an error message is shown for a split second, then the page reloads?

AltamashShaikh:
> Functionality works and the copy is being stopped. Only one issue, when the copy fails, an error message is shown for a split second, then the page reloads?

@lachiebol That's the general behaviour for other cases, I will check if we can do anything here.

lachiebol:
> > Functionality works and the copy is being stopped. Only one issue, when the copy fails, an error message is shown for a split second, then the page reloads?
> 
> @lachiebol That's the general behaviour for other cases, I will check if we can do anything here.

@AltamashShaikh If it's normal behaviour and the fix isn't quick, happy to approve

AltamashShaikh:
@lachiebol Can you check again, need to handle this via try/catch block, otherwise core read the xhr status and does redirect if status is 401.

lachiebol (dismissed):
That fix is working great

AltamashShaikh:
@lachiebol Added similar check for triggers and variables too :+1: 
We don't have any triggers hidden behind use_custom_template capability, but you can test it with `Custom JavaScript` variable.

lachiebol (approved):
Looks good to me :+1:

PR #1078, opened 2026-01-15
Fixes escaping in container selector tooltip

AltamashShaikh:
## Description
Fixes escaping in container selector tooltip

1. 
2. 
3. 

## Checklist
- [NA] Version bumped?


sgiehl (approved):
Tested locally. Works 👍

PR #1082, opened 2026-02-25
Updated API documentation, #PG-5005

lachiebol:
1. 
2. 
3. 

## Checklist
- [✔] Version bumped?

### DIFF
--- a/API.php
+++ b/API.php
@@ -409,7 +409,7 @@ class API extends \Piwik\Plugin\API
      *
      * Note: If the current site already has a default container, another default container will be created.
      *
-     * @param int $idSite
+     * @param int $idSite The id of the site that the container will be created for
      * @return string The ID of the created container.
      */
     public function createDefaultContainerForSite($idSite)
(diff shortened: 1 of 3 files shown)


API.php
```diff
@@ -786,6 +787,7 @@ public function addContainerTrigger($idSite, $idContainer, $idContainerVersion,
      *                            array(array('actual' => 'VARIABLENAME', 'comparison' => 'equals', 'expected' => 'expectedValue'))
      *                           To get a list of available comparisons, call {@link TagManager.getAvailableComparisons}
      * @param null|string $description   Optional description
+     * @return null
```
lachiebol, line 790: This doesn't feel correct? In the method we assign a $return variable that's always null. Is there a reason? Or could we fix and just return void?

AltamashShaikh: @lachiebol This is wrong, it should ideally return the $idTrigger, this is consistent for tag and variables too.

AltamashShaikh (approved):

PR #1085, opened 2026-03-03
Adds restriction for tag and triggers actions based on customTemplate capability, #AS-504

AltamashShaikh:
## Description
Adds restriction for tag and triggers actions based on customTemplate capability,

## Issue No
#AS-504

1. 
2. 
3. 

## Checklist
- [NA] Version bumped?


API.php
```diff
@@ -1070,7 +1084,11 @@ public function deleteContainerVariable($idSite, $idContainer, $idContainerVersi
         $this->accessValidator->checkWriteCapability($idSite);
         $this->containers->checkContainerVersionExists($idSite, $idContainer, $idContainerVersion);
 
-        if ($this->getContainerVariable($idSite, $idContainer, $idContainerVersion, $idVariable)) {
+        $variable = $this->getContainerVariable($idSite, $idContainer, $idContainerVersion, $idVariable);
+        if ($this->variablesProvider->isCustomTemplate($variable['type'])) {
```
lachiebol: Should we check `$variable` is not empty before checking `'type'`?

lachiebol (changes requested):
LGTM :) Buttons not showing now in UI and API is blocking, just one comment around an empty value check

API.php
AltamashShaikh: Yes, updated :+1:

lachiebol (dismissed):

lachiebol (approved):

PR #1103, opened 2026-04-27
PG-3694 tag manager navbar

brenthoneybone:
## Description
Tag manager menu isn't consistent with the rest of the product. This implements the updated nav, with containers down the left nav and a search input more inline with the rest of the product e.g. funnels/goals etc

## Issue No
PG-3694

1. Click tag manager in the top menu
2. Container dropdown in top menu should be removed
3. Left menu should contain containers + sub menus in line with the rest of the product

## Checklist
- [✔/✖/NA] Version bumped?
- [✔/✖/NA] Documentation updated?


lachiebol (dismissed):
LGTM functionality wise, nice to have everything consistent.

One nitpick/suggestion but not blocking, could we check with Stan if there's a nicer way to display the selected container? It looks nice with multiple containers but looks a bit strange with a single container:

Image

Maybe something like this:

Image

Feel free to ignore if Stan is already happy with it 😄

AltamashShaikh (changes requested):
@brenthoneybone the dropdown now introduces a security concern, earlier creating a container with name "Test  Imm" escapes the text, but now the xss is executed, you need to escape the name.

Image

AltamashShaikh:
@brenthoneybone Review from codex
```
High-risk issues (blocking)

  - Container names now share the same submenu keyspace as built-in Tag Manager actions, so a perfectly valid container name can overwrite menu entries. addContainerMenuItems() inserts containers
    under TagManager_TagManager using the container name as the submenu key (Menu.php:128), and configureContainerSubMenu() later inserts entries like Dashboard, TagManager_Tags, etc. into that same
    bucket (Menu.php:91). Since MenuAbstract stores submenu items by name key (/var/www/html/work/matomo/core/Menu/MenuAbstract.php:229), a container named Dashboard, TagManager_Tags, or any other
    colliding label will replace or be replaced by a real action. That makes the sidebar non-deterministic for legitimate user input.

  Medium-risk issues

  - The custom sidebar template dropped the parent active-state behavior that CoreHome’s menu had. In the stock template, menuTab gets active when any child is active (/var/www/html/work/matomo/
    plugins/CoreHome/templates/_menu.twig:28); in the new templates/_menu.twig:72, the top-level <li class="menuTab"> is never marked active. That is likely to regress visual highlighting and may
    affect any CSS/JS that depends on the parent tab being active
```

brenthoneybone:
@AltamashShaikh nice catches. I have fixed the xss vulnerability and added a test around it so it is checked. Also patched the two other codex recommendations.

lachiebol (changes requested):
Fixes mostly LGTM 👍 

Only thing I could find is that we're still rendering HTML via the title

Image

AltamashShaikh (changes requested):
@brenthoneybone Can still see xss being executed

brenthoneybone:
image
image
image

Should be good now in all locations (single container, multiple container and mobile)

AltamashShaikh (commented):
functionally looks good to me, @lachiebol Can you also check once ?

lachiebol (approved):
XSS is fixed, LGTM 👍

AltamashShaikh (approved):

AltamashShaikh:
@brenthoneybone Do you want to wait for Jeff to re-check or should we merge this one ?

brenthoneybone:
I think it's ok to merge. I've also sent a demo video to Stan and he was happy. I'll update Jeff and we can add a new ticket for any additional changes/findings for the next release. It seems like you two also checked it out and seem to not have found anything major that breaks?

PR #1117, opened 2026-05-17
Adding fix for state leaks

chippison:
## Description
The tests started to fail once we split more of the integration tests to different runners. 
This is a small change to fix a failing test because of some state leaks from previous tests

1. 
2. 
3. 

## Checklist
- [NA] Version bumped?
- [NA] Documentation updated?

### DIFF
--- a/tests/Integration/Context/WebContextTest.php
+++ b/tests/Integration/Context/WebContextTest.php
@@ -52,7 +52,7 @@ class WebContextTest extends IntegrationTestCase
 
     public function tearDown(): void
     {
-        TagManager::$enableAutoContainerCreation = false;
+        TagManager::$enableAutoContainerCreation = true;
         parent::tearDown();
     }
 
--- a/tests/Integration/Input/AccessValidatorTest.php
+++ b/tests/Integration/Input/AccessValidatorTest.php
@@ -13,6 +13,7 @@ use Piwik\Plugins\TagManager\Access\Capability\PublishLiveContainer;
 use Piwik\Plugins\TagManager\Access\Capability\TagManagerWrite;
 use Piwik\Plugins\TagManager\Input\AccessValidator;
 use Piwik\Plugins\TagManager\SystemSettings;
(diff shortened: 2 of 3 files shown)


tests/Integration/Context/WebContextTest.php
```diff
@@ -52,7 +52,7 @@ public function setUp(): void
 
     public function tearDown(): void
     {
-        TagManager::$enableAutoContainerCreation = false;
```
chippison, line 55: Since this was set to `false` on setup, the assumption is to set this back to `true` on teardown?

tests/Integration/Input/AccessValidatorTest.php
```diff
@@ -45,6 +46,8 @@ public function setUp(): void
     {
         parent::setUp();
 
+        TagManager::$enableAutoContainerCreation = false;
+
```
chippison, line 50: Since the previous test above has always set `$enableAutoContainerCreation` on teardown, this was also false when it reached this test when they were running on the same process. 

Now that we have split the tests more into more processes, the error starts showing up, so we added the code above on this suites setup and teardown

chippison:
Also, the failing UI tests are still related to the the PR: `PG-5068-change-inline-helptexts`

tests/Integration/Context/WebContextTest.php
lachiebol: That seems correct to me, reset to the default value

lachiebol (changes requested):
Fix LGTM 👍 

Can we check if this test needs the fix as well?

https://github.com/matomo-org/tag-manager/blob/4f905213b5c507abfd90f400afc658d0e23ef6be/tests/Integration/Template/Tag/MatomoTagTest.php#L36

lachiebol (approved):

lachiebol:
@chippison Want me to merge this one now?

chippison:
@lachiebol Yes please

PR #1123, opened 2026-06-08
Hardens the copy code based on dependent triggers variable type, #AS-555

AltamashShaikh:
## Description
Hardens the copy code based on dependent triggers variable type

## Issue No
#AS-555

1. 
2. 
3. 

## Checklist
- [NA] Version bumped?
- [NA] Documentation updated?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 tests/System/processed/*xml
 tests/System/processed/*text
+.codex
 /vue/dist/demo.html
 /vue/dist/*.common.js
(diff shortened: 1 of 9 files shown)


.codex
lachiebol, line 1: Is this supposed to be git ignored?

lachiebol (changes requested):
Most cases LGTM, just the param order issue mentioned in DM might need to be addressed

lachiebol (approved):
LGTM

PR #1134, opened 2026-06-30
Improves container delete to check if its released and has permissions, #AS-593

AltamashShaikh:
## Description
Improves container delete to check if its released and has permissions

## Issue No
#AS-593

1. 
2. 
3. 

## Checklist
- [NA] Version bumped?
- [NA] Documentation updated?


lachiebol (approved):
Looks good, we don't need to update the UI to reflect this?

AltamashShaikh:
> Looks good, we don't need to update the UI to reflect this?

@lachiebol Should be fine, the users will get an UI error.

PR #1136, opened 2026-07-02
Adds code to escape measurementID in GA tag

AltamashShaikh:
## Description
Adds code to escape measurementID in GA tag

1. 
2. 
3. 

## Checklist
- [NA] Version bumped?
- [NA] Documentation updated?


snake14 (approved):
Looks good to me. The only thing that I can think of is maybe adding a test case around the change.

PR #1144, opened 2026-07-24
Adds code to tighten the check for publish capability user, #PG-5347

AltamashShaikh:
## Description
Adds code to tighten the check for publish capability user

## Issue No
#PG-5347

1. Disallow CustomHTML tags creation if user has only published capability
2. The capability would still allow to create a new version with custom tags

## Checklist
- [NA] Version bumped?
- [NA] Documentation updated?

### DIFF
--- a/API.php
+++ b/API.php
@@ -1187,9 +1187,9 @@ class API extends \Piwik\Plugin\API
         }
 
         $this->enableGeneratePreview = false;
-        $container = $this->accessValidator->runWithoutCustomTemplatesCapabilityCheck(
-            fn () => $this->containers->createContainerVersion($idSite, $idContainer, $idContainerVersion, $name, $description)
-        );
+        $container = $this->accessValidator->runWithoutCustomTemplatesCapabilityCheck(function () use ($idSite, $idContainer, $idContainerVersion, $name, $description) {
+            return $this->containers->createContainerVersion($idSite, $idContainer, $idContainerVersion, $name, $description);
+        });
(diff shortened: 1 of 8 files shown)


snake14 (commented):
Looks good to me.

AI listed the following:

> Medium
> 
>   1. Cross-site container copies can leave an orphaned destination container
>      when custom-template access is denied. The new unconditional check in /tmp/
>      tag-manager-pr1144/API/Import.php:101 runs during container copying, but /
>      tmp/tag-manager-pr1144/Controller.php:489 checks custom-template capability
>      only for the source site. /tmp/tag-manager-pr1144/Model/Container.php:520
>      inserts the destination container before importing its contents at line
>      535. A user with custom-template access on the source and write/publish
>      access—but no custom-template access—on the destination therefore receives
>      an error while the empty container remains. Preflight the destination
>      capability before insertion or roll back the newly created container when
>      import fails, and add a cross-site regression test.
> 
>   Low / Polish
> 
>   1. The privileged bypass added at /tmp/tag-manager-pr1144/API.php:1513 lacks
>      targeted rollback coverage. The existing rollback test runs as superuser,
>      so it would still pass if this wrapper were removed. Add a publisher-
>      without-custom-capability test whose existing draft contains custom
>      content, whose replacement import fails after mutation, and which verifies
>      both draft restoration and subsequent capability enforcement.

AltamashShaikh:
@snake14 Fixed, please check.

snake14 (approved):
Nice work. 👍

PR #1153, opened 2026-07-31
Clarify the preview mode notification (PG-3551)

werner-calitz-innocraft:
## Description

While preview mode is enabled, the notification at the top of every container screen offered two competing actions and a paragraph telling users to append `?mtmPreviewMode` / `&mtmPreviewMode` to their site URL. Once a debug URL was entered those two strings became links, and clicking one navigated the user out of Matomo **in the same tab**, hiding their configuration.

The links opened in the same tab because of a sanitiser interaction rather than a missing attribute. The template hand-built the anchor:

```twig
{% set mtmPreviewIdUrlStart = '<a href="' ~ previewUrl|e('html_attr') ~ '" target="_blank">' %}
```

Raw notification HTML is re-rendered through `v-html="$sanitize(message)"`, and `$sanitize` adds this hook in `plugins/CoreVue/polyfills/src/index.ts`:

```ts
DOMPurify.addHook('afterSanitizeAttributes', (node) => {
  if (node.hasAttribute('target') && node.getAttribute('target') === '_blank'
    && (!node.hasAttribute('rel') || !hasSafeRel(node.getAttribute('rel')))) {
    node.removeAttribute('target');
  }
});
```

The anchor carried no `rel`, so `target="_blank"` was stripped. This is also why the FAQ link on the same notification *did* open a new tab — it was built with `externallink()`, which emits `rel="noreferrer noopener"`.

Rather than only fixing the attribute, the component is restructured around a single clear call to action:

- Title plus a secondary outlined **Disable Preview** button in the top right.
- Copy explaining what preview mode actually does, ending in a *Learn more* link to the [preview/debug FAQ](https://matomo.org/faq/tag-manager/preview-debug-a-tag-manager-container/), built with `externallink()`.
- The URL input and **Debug** button moved to the bottom, leaving one primary CTA.
- The share-preview and "Debug not working?" paragraphs dropped, removing the problem links entirely.

Supporting changes:

- **Accessibility** — Disable Preview was an `<a>` with no `href`, so it was not focusable or keyboard-operable. It is now a `<button type="button">`, which matters more now that it is styled as a button. The delegated handler selector moved with it.
- **Styling** — moved out of `manageEdit.less` into its own `stylesheets/previewDebugNotification.less` using BEM naming, replacing a hardcoded `#ccc` with `@theme-color-border-alternative` so the input border is theme-aware. `Disable Preview` uses `btn btn-outline`, one of the three button variants Morpheus documents.
- **Strings** — three new keys; the three now-dead `PreviewDebugEnabledNotificationLine*` keys removed. English only, since the plugin is Weblate-managed.
- **Cleanup** — `Controller.php` no longer computes `$mtmPreviewId` / `$previewUrl`, and the then-unused `SafeDecodeLabel` import is dropped (PHPCS enables `SlevomatCodingStandard.Namespaces.UnusedUses`).

## Issue No

PG-3551

## Checklist
- [NA] Version bumped?
- [NA] Documentation updated?


snake14 (dismissed):
Looks good! 👍 

Optional: I noticed a couple styling issues while testing locally -

1. The alignment of the input and button is a little odd: 
Image

2. The tabs in debug mode are squished together and don't have clear division: 
Image

snake14 (approved):
Looks good. 👍 

AI had one non-blocking comment:

> ### Medium
> 
>   1. The new regression test does not verify the sensitive URL parameters are absent. tests/UI/TagManager_spec.js:163-167 checks only innerText, while the old implementation stored mtmPreviewMode and mtmSetDebugFlag in an anchor’s href. The test would pass against the old code. Assert the notification’s links/HTML do not contain those parameters.

PR #1157, opened 2026-08-04
Adds code to publish default container as a new version, #PG-5393

AltamashShaikh:
## Description
Adds code to publish default container as a new version, #PG-5393

## Issue No
#PG-5393

1. Create a new website
2. It will create a default container, but the draft and liver version will point to same version
3. Login as a write user and try creating any new tags and it will be denied
4. Checkout this PR and repeat step 1-3 and it will work fine
5. For old default version, we need to publish them, for tag creation to work for write users.

## Checklist
- [NA] Version bumped?
- [NA] Documentation updated?


snake14 (dismissed):
LGTM

AI just called out the failing CI items:

> Blocking
> 
>   1. The PR leaves the integration suite failing on both supported PHP jobs.
>      tests/Integration/APITest.php:527-536 selects $container['versions'][0]
>      after default-container creation, then attempts to edit it. Because
>      API.php:473-483 now publishes the newly created version, that version is
>      live-released and editing requires tagmanager_publish_live_container; CI
>      reports NoAccessException. Update the test to use $container['draft']
>      ['idcontainerversion'], then rerun the suite.

snake14 (commented):
LGTM

AI had some feedback:

> Reviewed and ran locally: full `plugins/TagManager/tests/Integration/APITest.php` passes (161 tests), and I confirmed both new behaviours are genuinely regression-tested by reverting `API.php` and `API/Import.php` to `5.x-dev` — the two new tests fail there as expected.
> 
> **Blocking**
> 
> None.
> 
> **Medium**
> 
> 1. **The new import tests don't exercise the paths they look like they do.** `checkReplacingEntitiesIsPossible()` has three loops (tags, triggers, variables), each returning on the first match. Container 1's draft already holds two `CustomHtml` tags from `TagManagerFixture`, so the check always resolves in the tags loop — the `CustomJsFunction` variable both tests add isn't what drives the outcome, and the triggers and variables loops have no coverage. Confirmed by deleting those two loops: all seven `test_importContainerVersion*` tests still pass. Could we assert each entity type against a draft containing only that type? `tests/Integration/API/ImportTest.php` looks like the easier home, and already covers the equivalent check for `checkImportContainerIsPossible()`.
> 
> 2. **No non-regression test for the ordinary case.** Nothing asserts that a user with `tagmanager_write` + `tagmanager_publish_live_container` but no `tagmanager_use_custom_templates` can still import a plain payload into a draft with no custom templates. I verified by hand that it works, so no bug — but a later tightening of this check could take importing away from publishers entirely with nothing failing in CI. A short positive test would lock the boundary down.
> 
> 3. **Existing containers — question.** The `createDefaultContainerForSite()` fix applies to new containers only, and step 5 of the description notes old default versions still need publishing for write-user tag creation. There's no `Updates/*.php` here — separate follow-up ticket, documented manual step, or in scope for this PR? Asking mainly so it's recorded: as it stands, every site created before this ships keeps the behaviour where write users can't edit the draft.
> 
> **Low / Polish**
> 
> 1. `API/Import.php` now carries six near-identical provider loops across `checkImportContainerIsPossible()` and `checkReplacingEntitiesIsPossible()`. Consistent with the file as it stands, so no change needed here — just a note that a single loop over provider/entity pairs would be the tidier shape if that file is revisited.

snake14:
@AltamashShaikh LGTM

AI had the following:

> ### Blocking
> 
>   1. An unauthorized import can still mutate the draft. API/Import.php:159
>      throws inside API.php:1505-1518, where all failures trigger rollback. The
>      rollback deletes and recreates entities under
>      runWithoutCustomTemplatesCapabilityCheck, changing their IDs; with
>      backupName, cleanup at API.php:1511 can fail the same authorization check
>      and leave the backup version behind. Perform this permission check before
>      entering the backup/rollback flow. This violates the security access-
>      control expectation.
> 
>   ### Medium
> 
>   1. Existing default containers remain unfixed. The change only affects newly
>      created sites, while plugin activation returns when a container already
>      exists (TagManager.php:222-226); no update migration repairs older
>      containers. Those sites still leave write users unable to edit the draft.
> 
>   2. Import coverage short-circuits through the tag path: the fixture draft
>      already contains a CustomHtml tag (tests/Fixtures/
>      TagManagerFixture.php:145), so the new variable test never exercises the
>      trigger or variable loops in API/Import.php:129-147. There is also no
>      positive Publisher-only import test for a draft without custom templates.
> 
>   ### Low / Polish
> 
>   None.

AltamashShaikh:
@snake14 Please check again.

snake14 (approved):
LGTM

Optional - AI had the following:

> ### Blocking
> 
>   None.
> 
>   ### Medium
> 
>   - Existing default containers are not repaired. The fix only applies during
>     new container creation (API.php:473-483); existing installations still
>     retain the old draft/live version because activation exits when a container
>     exists (TagManager.php:222-226). Add an idempotent update migration or
>     equivalent upgrade-time repair.
> 
>   ### Low / Polish
> 
>   - No regression test covers the custom-trigger path. Tests cover custom tags
>     and variables, but not the trigger checks in API/Import.php:146-150 and
>     API.php:1576-1579.

PR #1158, opened 2026-08-04
Update the view-access-restricted screenshot for the always-enabled s…

chippison:
## Description
Update the view-access-restricted screenshot for the always-enabled sign in button

That page renders the login form, whose submit button is no longer disabled while the fields are empty, so this baseline captured the old grey button.

## Issue No
DEV-20463

## Checklist
- [NA] Version bumped?
- [NA] Documentation updated?


snake14 (approved):
LGTM

Do all the failing screenshots need to be updated?

chippison:
Nah, I don't think those screenshots need to be updated on this PR. I think this is good to be merged

PR #1159, opened 2026-08-05
Update the view-access-restricted screenshot for the refreshed login layout

chippison:
## Description
Updates the view-access-restricted baseline for the refreshed login layout in matomo-org/matomo#24998. That page renders the sign-in form with an error notification, and #24998 replaces the centred-card-under-a-top-bar shell with the refreshed single column, so the previous baseline still captured the old chrome.

Cut from `prepare6x` rather than stacked on #1158, so it can be reviewed and merged on its own. The PNG is replaced wholesale, so it does not depend on #1158's intermediate update. When #1158 merges into `prepare6x` this branch will need a rebase, keeping this version of the baseline.

The PNG is the `/processed/` output of Matomo core CI run 30970737938, verified byte-identical by SHA-256 against the artifact rather than rendered locally.

Heads up on CI: the `UI` check here will stay red until matomo-org/matomo#24998 merges, because the baseline describes login markup that is not yet in the core branch this plugin's CI builds against. #1158 is currently red for exactly the same reason.

## Issue No
DEV-20463

## Checklist
- [NA] Version bumped?
- [NA] Documentation updated?


AltamashShaikh (dismissed):
LGTM

chippison:
Should be good to be reviewed and merged into `prepare6x`

AltamashShaikh (commented):
@chippison There are other UI screenshots which are failing too, unrelated to this one.

tzi:
ℹ️ This PR has a lot of conflicts to solve

chippison:
We can close this off as this has been fixed in: 
https://github.com/matomo-org/tag-manager/pull/1161/changes#diff-2ee09ccc46cba2c8bbf15398fbfe8c6e612eb6a5f37030f3fac6b41440e38f3b

PR #1163, opened 2026-08-09
Prepare TagManager for Matomo 6

sgiehl:
### Description

Matomo 6 compatibility for TagManager, targeting the new `6.x-dev` maintenance branch.

`6.x-dev` was created at the `5.x-dev` tip, so it carries all current Matomo 5 work; the existing `prepare6x` was rebased onto it.

### What changed
- **CI** — `matomo6_min_php` / `matomo6_max_php` aliases replace pinned literals across the matrix, both single-version jobs, the `upload-artifacts` condition and the header comment, so the workflow follows core's supported range instead of drifting.
- **PHPStan** — `phpVersion: 80100` and `excludePaths` pruned. PHPStan 2 rejects a literal exclude matching nothing, so **the old config aborted before analysing anything**: `vendor/` removed (no `composer.json`, no committed `vendor/`), `github-action-tests/` marked `(?)` (absent from the repo but real in CI).
- **Already on the branch** (rebased): PHP 8.1/8.5 matrix, MySQL 8.0 / MariaDB 10.6, UI on Node 24, the Vite Vue rebuild with ESLint fixes, unminified UMD removal, `_glob()` → native `glob()`, refreshed screenshots.

### Worth a look
- **No version marker to set.** TagManager ships no `plugin.json` and no `composer.json` — it is versioned with core — so there is no `6.0.0` bump and no `require.matomo` to raise. `Updates/` tops out at `5.2.2-b1`, safely below core's `6.0.0-b1`.
- The rebase hit a conflict on four expected screenshots between `5.x-dev`'s #1145 and the Node 24 / Puppeteer 24 re-render in #1141. The Matomo 6 renders were kept; later commits re-render those same files anyway.
- **PRs #1161 and #1159 are based on the pre-rebase `prepare6x` and will need rebasing.**
- PHPStan's 25 remaining level-1 findings are pre-existing and unrelated; this repo runs no PHPStan workflow.

⚠️ Separately: `6.x-dev` is **8 non-test commits behind `5.x-dev`** (AS-643 fixes, default-container publish, translations). Needs a backport — not part of this PR.

### Checklist


snake14 (approved):
Looks good to me. 👍

PR #1164, opened 2026-08-10
Update the general settings screenshots for the TrackingSpamPrevention core integration

sgiehl:
## Description

Updates the two general settings screenshots for the `TrackingSpamPrevention` core integration in matomo-org/matomo#25035.

With Matomo 6 that plugin is bundled with core and activated by default, so on `?module=CoreAdminHome&action=generalSettings` its settings card moves up in the plugin load order and shows the settings it gained since the previously pinned commit. Both screenshots in this plugin capture that page, so they change even though nothing in Tag Manager itself does.

Both images are taken from the matomo-org/matomo#25035 build artifacts, so they are exactly what CI renders, and this branch changes nothing besides them.

This has to be merged before matomo-org/matomo#25035, which pins the submodule to this branch.

## Issue No

Related to matomo-org/matomo#25035

1. Check out matomo-org/matomo#25035, which bundles `TrackingSpamPrevention` with core.
2. Run the Tag Manager UI tests. Expected: they pass. Actual: `TagManager_settings_page` and `TagManager_update_restrict_setting` differ, because the general settings page now renders the plugin's settings card in a different position and with more settings.

## Checklist
- [NA] Version bumped?
- [NA] Documentation updated?

### DIFF
--- a/tests/UI/expected-screenshots/TagManager_settings_page.png
+++ b/tests/UI/expected-screenshots/TagManager_settings_page.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:e3e3d6ec8ba860c1d68610fbcacecf828559a45a0f937207603d4429b0c48a84
-size 115343
+oid sha256:5875c21e17e5c7270da90a0583a950297b0dbb1c43eb61b026ca4b0e5ca58858
+size 115342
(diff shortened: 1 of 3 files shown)


snake14 (approved):

snake14:
I'll wait to merge this PR until the parent PR is merged.

PR #1165, opened 2026-08-11
Follow the redesigned site selector dropdown in the UI tests

tzi:
### Description

Points the two site selector UI tests at the redesigned dropdown panel and regenerates the six screenshots that capture it.

### Impact

None visible to users: this repository only changes UI test selectors and expected screenshots, both of which follow the site selector redesign in core.

#### With the core branch pinned

1. Check out matomo-org/matomo#24987 (`design-system-site-selector-dropdown`), which pins this branch as `plugins/TagManager`.
2. Run `ddev matomo:console tests:run-ui --plugin=TagManager`.
  👁️ All specs pass, including the six site selector screenshots.
3. Go to *Tag Manager → Manage Containers* and use the *Copy* action on a container.
  👁️ The site selector panel matches `Container_copy_container_site_select.png`: 254px wide, with the design system search field at the top.

#### Why the selector changed

1. Open `tests/UI/Container_spec.js` and `tests/UI/TagManager_spec.js`.
  👁️ Both now query `.top_bar_sites_selector .piwikSelector__dropdown` instead of `.top_bar_sites_selector .dropdown`.
  👁️ The core PR drops the `dropdown` class from the panel, so the old selector matches nothing and the specs would capture an empty element rather than fail loudly.

#### Against core 6.x-dev

1. Run the same suite with core on `6.x-dev` instead.
  👁️ Expect the six screenshots to fail: they describe the redesigned panel, so they only pass together with the core PR.

### Checklist


snake14 (commented):
Thank you @tzi :+1:

snake14 (approved):

PR #1175, opened 2026-08-18
Update expected UI screenshots for the restyled select [ignore_release]

snake14:
## Description

Updates the expected UI screenshots for the restyled select form field in matomo-org/matomo#25100, which brings `uicontrol="select"` in line with the expandable selects and the top-level selectors: a bordered, rounded control with a chevron instead of Materialize's underlined field.

Screenshots only — no plugin code changes, so no version bump or changelog entry. The images are the processed output of the core PR's own CI run, not locally generated.

**Depends on matomo-org/matomo#25100.** The UI tests here run against core `6.x-dev`, which does not have the restyle yet, so this PR's UI check stays red until that merges. Re-running it afterwards should turn it green with no further changes.

## Issue No
Related to matomo-org/matomo#25100

1. Run the plugin's UI tests against core `6.x-dev` with matomo-org/matomo#25100 applied.
2. Expected: all specs pass.
3. Actual (without these updated screenshots): the specs listed in this diff fail on the restyled select, with the old underlined control in the expectation.

## Checklist
- [NA] Version bumped?
- [NA] Documentation updated?


snake14:
Superseded by #1179, which regenerated the affected screenshots against the merged core restyle (matomo-org/matomo#25100). The remaining baselines this branch touched are correct as they stand on 6.x-dev. Closing.

PR #1177, opened 2026-08-19
Stop modal animation before capturing the delete-website confirmation

snake14:
## Description

`should show the container detail when delete button is pressed` fails on `6.x-dev` with an 8911 px difference. This started as a screenshot resync, on the assumption that #1161 had missed a capture after core [#25033](https://github.com/matomo-org/matomo/pull/25033). **That was wrong**, and the first version of this PR is reverted here.

Resyncing the baseline produced *exactly* the same 8911 px difference on the next run. Downloading both images and diffing them shows why: they are identical except for a soft gradient in a 742×70 band across the bottom — the Materialize modal footer's scroll shadow, captured part-way through its transition. The capture is non-deterministic, so no baseline can ever match it.

Waiting on the page does not help, because a CSS transition keeps repainting after the DOM has stopped changing: height and markup are already final while the shadow is still fading. So this stops animation outright before capturing, which is the same class of workaround `capture.js` already applies a few lines below, where Materialize's open animation does not advance under the newer headless Chrome and the modal has to be forced to its final position.

The helper is applied only in `capture.modal`, so nothing else changes its behaviour. The baseline is left exactly as it is on `6.x-dev`.

This cannot be reproduced locally without the UI harness, so this PR's own run is the verification. If the difference disappears, the capture was animation; if it comes back at 8911 px again, something other than the shadow is moving and the diff images are worth another look.

## Issue No

No separate issue — found while auditing the red `6.x-dev` builds ahead of Matomo 6.

1. Run the UI tests against Matomo `6.x-dev`.
2. Expected: `should show the container detail when delete button is pressed` passes consistently.
3. Actual: it fails at 8911 px, and still fails at exactly 8911 px after the baseline is resynced.

## Checklist
- [NA] Version bumped?
- [NA] Documentation updated?

### DIFF
--- a/tests/UI/capture.js
+++ b/tests/UI/capture.js
@@ -5,6 +5,15 @@
  * @license http://www.gnu.org/licenses/gpl-3.0.html GPL v3 or later
  */
 
+// A CSS transition keeps repainting after the DOM has stopped changing, so waiting on the page cannot tell
+// whether the pixels have settled. The modal footer's scroll shadow is the one that bites: it is still fading
+// when the capture lands, which differs by a few thousand pixels between otherwise identical runs.
+exports.disableAnimations = async function (page) {
+  await page.webpage.addStyleTag({
+    content: '*, *::before, *::after { transition: none !important; animation: none !important; }',
+  });
+};
+
 // the first table row can for some reason can have height that varies randomly by 1px.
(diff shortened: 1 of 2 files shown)


werner-calitz-innocraft (approved):
LGTM

PR #1184, opened 2026-09-02
Use the shared loader component instead of the loading GIF

chippison:
DEV-20697

### The bug

`plugins/Morpheus/images/loading-blue.gif` has no alpha channel — every pixel is opaque, over half of them pure white. So it draws a white block instead of a spinner: a white rectangle in dark mode, and near-invisible bars on a light background.

An `` renders that opaque white exactly the same way a CSS background does, so every one of these call sites has the bug.

### The fix

Swap the `` for the shared loader component core already exports. Same three bars, same 16×11 size, and it takes its colour from the surrounding text so it reads in both themes.

No new markup or CSS — core already styles `.loadingPiwik img` and `.loadingPiwik .matomo-loader` identically, so the swap is visually neutral.

| Before | After |
| --- | --- |
| Screenshot 2026-09-03 at 2 47 02 PM | Screenshot 2026-09-03 at 2 47 38 PM |

### Checklist

[✔] I have understood, reviewed, and tested all AI outputs before use
[✔] All AI instructions respect security, IP, and privacy rules


snake14 (dismissed):
LGTM 👍 

Please check a couple non-blocking items AI listed:

> - Low — List views say "Loading data" while updating. It's pre-existing, but the branch just rewrote those exact five lines and its own final commit is about distinguishing the two states, so leaving the lists inconsistent is a gap the PR itself opened up. One-line fix per file.
> - Low — the commit message's rationale is wrong for ContainerEdit/VersionEdit (their <form> has no v-show, so the states can overlap). The conclusion still holds, so it's a wording fix — but on a squash-merge that text becomes the permanent history, so it's worth a nudge.

snake14 (approved):
Thank you 👍

PR #1193, opened 2026-09-08
Follow the option list out of the modal [ignore_release]

snake14:
## Description

Core now renders an expandable select's option list at the page level instead of inside the field, so a modal or any other scrolling ancestor can no longer clip it (matomo-org/matomo#25225). Two things in Tag Manager still assumed the old arrangement.

`ContainerTag_spec.js` reached the list through `.modal.open`, which now matches nothing — and a jQuery click on an empty selection is a silent no-op, so the "level 2" test expanded nothing and captured the level-1 state. Both popup captures framed only the modal, which the list has since left, so the very thing those two shots exist to show is unclipped had dropped out of frame entirely.

`capture.modalWithOptionList()` now captures the modal together with the list, and the level-2 step clicks the list where it actually lives and waits for the second level to appear. The stylesheet's `position` override for the in-modal list goes with them, since it matches nothing once the core change lands.

Please merge this after matomo-org/matomo#25225 — until that lands, the removed rule is still doing its job.

The two `edit_trigger_directly_popup_list_level*` expected screenshots still need regenerating: the capture region changes by design.

## Issue No

Related to matomo-org/matomo#25225

1. With matomo-org/matomo#25225 applied, open a container, edit a tag, open the fire-trigger edit modal and open the trigger type expandable select.
2. Expected: the option list is fully visible in the capture, and the level-2 shot shows an expanded category.
3. Actual: the captures frame only the modal so the list is missing, and the level-2 step clicks an empty selection and captures the level-1 state.

## Checklist
- [✖] Version bumped?
- [NA] Documentation updated?


AltamashShaikh (approved):
LGTM

AltamashShaikh:
@snake14 You would need to update the failing screenshot.

snake14:
Thanks @AltamashShaikh. Those three are expected to fail here and the images are already correct: this repo's UI job runs against core `6.x-dev` (`matomo-test-branch: ${{ github.base_ref }}`), which doesn't yet have matomo-org/matomo#25225, so the option list still renders inside its field. Regenerating them against current core would make them wrong as soon as that lands. They pass in the core PR, which pins this branch — `UI-plugins (TagManager)` is green there — and this job goes green on a re-run once #25225 merges.

PR #1201, opened 2026-09-18
Raise PHPStan to level 5 and clear what it found [ignore_release]

snake14:
## Description

Raises the plugin's PHPStan level from 1 to 5 and clears the 110 findings that surfaced. No behaviour changes — nothing a user of Tag Manager will see differs.

Dead branches are gone: the `method_exists()` shim for `Request::getRootApiRequestMethod()`, the null checks on an `$idDestinationSite` that has just been defaulted, a write-capability test nested inside an identical outer one, and the guards around a `getLoadedPlugin()` call that throws rather than returning nothing. Nineteen closures captured variables they never used, and three API methods kept a return value the models do not produce.

The rest was docblocks disagreeing with the code: the template providers and `BaseTemplate` return null until they find a match, the parameter migrators take any scalar or array as a default, `Environment` holds the `SystemSettings` it is constructed with rather than a `Configuration`, `FixedIdGenerator` counts with an int, and `ContainerVersionsDao` documented a `$statuses` parameter none of its three methods take.

`phpstan.neon` pins what the analyser cannot see, each by identifier, path and count: the third `loadTemplate()` parameter the pre-configured variables add, the superglobals the preview loader restores when `variables_order` leaves them out, core's own `Controller::$site` and `getCurrentUserTokenAuth()` docblocks, and the `NoAccessException` that `Site`'s constructor raises through SitesManager's API. `Updates/5.2.0-b4.php` passes the tag type where `NewTagParameterMigrator::addField()` expects a field name, so its third argument is ignored; `Tag::updateParameters()` runs the result through `formatParameters()`, which keeps only the parameters the template declares, so the stray key is dropped and `areCustomDimensionsSticky` takes its template default either way. A released update has to keep running as it shipped, so that one is pinned rather than corrected.

## Issue No

Related to PG-4897

Not applicable — this raises static analysis coverage rather than fixing a defect.

## Checklist
- [✖] Version bumped?
- [NA] Documentation updated?


Context/BaseContext.php
```diff
@@ -345,7 +348,7 @@ protected function parameterToVariableJs($value, $container)
                 }
 
                 $pos = $end + $ignoreLengthOpeningBrackets;
-            } while ($end !== false);
+            } while (true);
```
AltamashShaikh, line 351: @snake14 Looking at the code this looks safe, can you check if this will not result in a never ending do,while loop ?

AltamashShaikh (approved):
left 1 comment to check, otherwise good to merge

Context/BaseContext.php
snake14: Yes. The loop is safe and exited successfully, one way or another, during all testing of the code.

